PR #2, opened 2022-08-29
Face detection example

erikbern:
We might want to changes this to use YOLO or something better at some point (generalizing it to object detection)

image


erikbern:
Merging this for now, but happy to improve this in the future

PR #10, opened 2022-09-02
Doc OCR example updates

aksh-at:
Updated the styling and added more screenshots:

image


erikbern:
beautiful!

ekzhang:
Hey could we avoid putting images bigger than a few hundred KB in the repository? the screenshots in this PR are several MB and this will quickly slow things down

aksh-at:
oops good point, I'll resize all of the screenshots in the docs directory.

PR #15, opened 2022-09-09
Adds a simple duckdb example

freider:
* Fetches stock tickers using  yfinance
* Transforms resulting data using pandas
* Use duckdb pandas integration to insert into duckdb
* Persist duckdb in modal shared volume for later querying
* Current querying is a simple window function (which I assume duckdb is good at...)

### DIFF
--- a/08_advanced/duckdb_example.py
+++ b/08_advanced/duckdb_example.py
@@ -1,0 +1,52 @@
+import modal
+
+stub = modal.Stub(image=modal.DebianSlim().pip_install(["duckdb", "pandas", "yfinance"]))
+shared_volume = modal.SharedVolume().persist("db_storage")
+
+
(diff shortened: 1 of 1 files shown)


08_advanced/duckdb_example.py
```diff
@@ -0,0 +1,60 @@
...
+            date,
+            variable AS ticker,
+            value AS price
+        FROM bootstrap_df
+        WHERE value IS NOT NULL
+    )"""
+    )
+
+
+@stub.function(shared_volumes={"/db": shared_volume})
```
erikbern, line 38: what do you think about making this a webhook that renders the stock prices using [altair](https://altair-viz.github.io/user_guide/display_frontends.html)? just a random idea

PR #16, opened 2022-09-13
Convert two examples to use __enter__

erikbern:
By putting a method on a class and defining an `__enter__` method, you can provide container-specific initialization code:

```python
class Model:
    def __enter__(self):
        self.model = pickle.load(open("model.pickle"))

    @stub.function
    def predict(self, x):
        return self.model.predict(x)
```


09_job_queues/doc_ocr_jobs.py
```diff
@@ -42,39 +42,43 @@
 # decorator, we set up a Modal [Function](/docs/reference/modal.Function) that uses GPUs,
 # has a [`SharedVolume`](/docs/guide/shared-volumes) mount, runs on a [custom container image](/docs/guide/custom-container),
 # and automatically [retries](/docs/guide/retries#function-retries) failures up to 3 times.
+# In order for the model to be loaded only once per container, we put the model in a class and use a custom `__enter__`
```
thundergolfer, line 45:
```suggestion
# In order for the model to be loaded only once per container, we put the model in a class and use a custom `__enter__`.
```

PR #17, opened 2022-09-18
Update deprecated image names

erikbern:


ekzhang (approved):
Looks good to me! Seems it would be useful to merge since our examples are showing deprecation warnings to users right now.

PR #24, opened 2022-09-21
Add <img/> alt tag to clear warning in frontend.

thundergolfer:
Clearing away the following warning:

image


ekzhang:
By the way it is better if we did this using the Markdown component, which gets styling applied by default.

`[alt text](./path_to_image.jpg)`

The video tag was a one-off thing.

PR #35, opened 2022-09-30
whisper demo

thundergolfer:
* Main thing I'm still not happy with is the post-submit flow, where a user will just be waiting there for like 10 minutes. Need to redirect them somewhere while they wait.

image


erikbern:
I think this has the potential to be a very compelling demo so maybe we should discuss how to get it there?

ekzhang:
Happy to collaborate on web work for this

PR #47, opened 2022-10-06
fix: screenshot example can now run > 1times on same machine

thundergolfer:
I think previously we were just exercising this example with cold AWS Lambda functions.


erikbern:
nice!

PR #51, opened 2022-10-10
Use DuckDB for the NYC taxi matrix

erikbern:
Turns out this is already in Parquet and so it's ready to go!


erikbern:
Merging it for now – will iterate more on it

PR #61, opened 2022-10-17
Whisper front-end updates

aksh-at:
https://modal-labs-whisper-pod-transcriber-fastapi-app.modal.run/


ekzhang:
Strong preference towards not having a `package{-lock}.json` for these examples and using a buildless system instead, since NPM packages go out of date quickly, they add a lot of continuous maintenance burden for a small web example

aksh-at:
Yeah this example was originally buildless, just like all the other ones we have. But it was kind of annoying to use the deps I wanted to use, and it seems good to have one example of using a builder for a not-that-small example.

ekzhang:
Makes sense!

PR #66, opened 2022-10-21
Add basic Dask example, with debugging/profiling code incl

thundergolfer:
* 1st commit includes the debugging and profiling I added when things were slow
* 2nd commit strips it out.


thundergolfer:
Probably won't bother merging this at this point.

PR #68, opened 2022-10-24
Add {fastai, weights&biases, gradio.app} demo, doing CIFAR-10 task

thundergolfer:
> **https://modal-labs-fastai-wandb-gradio-cifar10-demo-fastapi-app.modal.run/**

* Train a PyTorch vision model with `fastai`. Does transfer learning on GPU and takes about 40 minutes
* Model training is tracked with Weights & Biases
* Model is demo'd with Gradio.app

> Requires **`wandb`** secret which is not yet provided in Modal labs org. Dev'd this example in my personal account.
> _Edit:_ I've added `wandb` to our org, so team can run the example but you won't be able to view stuff in W&B because it's my personal key.

### Things learnt

* Unpacking the dataset `.tar.gz` into a shared volume was realllly slow. Ditched dataset caching between training runs as it was actually faster.
    * Would be cool if the deployed shared volume apps had observability in the Modal dashboard. eg. files written, stored size.
* It would be great to use `app.app_id` as the Weight & Biases run ID, but there's no way to access that in a detached app run.
* 1st-class CLI functionality would make detached runs nicer, and cut out a decent amount of ugly and rudimentary `sys.argv` code.

image


erikbern:
> * It would be great to use `app.app_id` as the Weight & Biases run ID, but there's no way to access that in a detached app run.

it's possible `modal.container_app.app_id` works, but I'm not sure

erikbern:
> Downloading is expectedly faster on Modal, but unpacking the 60,025 image files is significantly slower. Wonder if this is entirely explainable by differences in disk performance (AWS EBS vs. top-of-the-line SSD).

You mentioned this is on a shared volume (EFS) right? My guess is that EFS just isn't fast for lots of tiny operations – the EFS latency will dominate, and it's substantially higher than EBS (think about 10x – 1ms vs 100µs)

erikbern:
is the W&B dashboard public? would be cool to see!

thundergolfer:
> You mentioned this is on a shared volume

I originally tried a shared volume, but it was even slower. So this example now just downloads to local disk on the container (EBS).

> is the ...

Just made public: https://wandb.ai/thundergolfer/fast-cifar10/overview

PR #77, opened 2022-11-04
continuous deployment for (some) examples

thundergolfer:
Dry-run by default.

```
$ python3 example_deploy.py
INFO: dry-run is active. Intended deployments will be displayed to console.
⏩ skipping: 'generators.py' is not marked for deploy
⏩ skipping: 'get_started.py' is not marked for deploy
⏩ skipping: 'hello_world.py' is not marked for deploy
...
⏩ skipping: 'sentence_embeddings.py' is not marked for deploy
⏩ skipping: 'shell.py' is not marked for deploy
⏩ skipping: 'tqdm_progress_bar.py' is not marked for deploy
🌵  dry-run: 'vision_model_training.py' would have deployed
⏩ skipping: 'webcam.py' is not marked for deploy
⏩ skipping: 'webscraper.py' is not marked for deploy
```

Filtering the deployment set.

```
python3 example_deploy.py --filter cron
INFO: dry-run is active. Intended deployments will be displayed to console.
⏩ skipping: 'say_hello_cron.py' is not marked for deploy
```

Actually deploying an app, with filtering to restrict deployment set.

```
python3 example_deploy.py --filter vision --no-dry-run
⛴ deploying: 'vision_model_training.py' ...
✔️ deployed 'vision_model_training.py
```


.github/workflows/cd.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/setup-cached-python
+
+      - name: Install Modal client package
+        run: pip install modal-client
+
+      - name: Modal token set
+        run: |
+          modal token set --token-id ${{ secrets.MODAL_MODAL_LABS_TOKEN_ID }} --token-secret ${{ secrets.MODAL_MODAL_LABS_TOKEN_SECRET }}
```
erikbern, line 23: instead of setting the token, you can just pass the token as an environment variable (`MODAL_TOKEN_ID` and `MODAL_TOKEN_SECRET`)

thundergolfer: True. I just copied this from our other repo.

example_deploy.py
```diff
@@ -0,0 +1,81 @@
...
+    if not deployable:
+        print(f"⏩ skipping: '{module_with_stub.name}' is not marked for deploy")
+        return None
+
+    deploy_command = f"modal app deploy {module_with_stub.name}"
+    if dry_run:
+        print(f"🌵  dry-run: '{module_with_stub.name}' would have deployed")
+    else:
+        print(f"⛴ deploying: '{module_with_stub.name}' ...")
+        r = subprocess.run(shlex.split(deploy_command), cwd=module_with_stub.parent, capture_output=True)
```
erikbern, line 33: we should figure out a way to do this not through a subprocess but by importing modal directly and calling some function – seems useful for CI type stuff in general

thundergolfer: Yeh I did think of that. I kept it simple for now, but we should adopt an API based approach not shell out.

example_deploy.py
```diff
@@ -0,0 +1,81 @@
...
+    arguments = parser.parse_args()
+
+    if arguments.dry_run:
+        print("INFO: dry-run is active. Intended deployments will be displayed to console.")
+
+    example_modules = (ex for ex in get_examples() if ex.type == ExampleType.MODULE)
+    filter_pttrn = (r".*" + arguments.filter + r".*") if arguments.filter else None
+    results = [
+        deploy(
+            deployable=bool(ex_mod.metadata.get("deploy")),
```
erikbern, line 65: we should probably just parse the metadata directly in `example_utils` and write it as attributes to the dataclass, but that's not necessarily within scope for this PR

PR #87, opened 2022-11-05
Improve reliability of whisper demo

thundergolfer:
* **Set a user agent**. Failing to download the audio with 403 is probably the most common error. I had someone message me about the demo not working and I'm betting this was the reason.
* **Handle errors in the UI**. Previously the UI would just get into a loop showing 0 containers running and never show the user that the transcription had failed. That's fixed now, and how it looks is shown below:

new-error-handling

---

I don't offer a nice way for a user to proceed after getting an error, but this is good enough for now.

### DIFF
--- a/misc/whisper_pod_transcriber/api.py
+++ b/misc/whisper_pod_transcriber/api.py
@@ -1,4 +1,5 @@
 import json
+import urllib
 from typing import List
 
 from fastapi import FastAPI, Request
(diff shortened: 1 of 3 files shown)


misc/whisper_pod_transcriber/api.py
```diff
@@ -92,12 +92,24 @@ async def transcribe_job(podcast_id: str, episode_id: str):
...
     graph: List[InputInfo] = function_call.get_call_graph()
 
+    try:
+        function_call.get(timeout=0.1)
+    except TimeoutError:
+        pass
+    except modal.exception.RemoteError as exc:
+        if exc.args:
+            inner_exc = exc.args[0]
+            if "HTTPError 403" in inner_exc:
```
thundergolfer, line 109: There's gotta be a better way to do this.

erikbern: yeah, this is weird, we should throw the original exception here, not an internal modal exception. seems like a bug.

PR #89, opened 2022-11-06
fix: don't leave failed transcription jobs in the 'in-progress' modal…

thundergolfer:
Working to address [reddit.com/r/MachineLearning/comments/ynz4m1/comment/ivc3w56](https://www.reddit.com/r/MachineLearning/comments/ynz4m1/comment/ivc3w56). Transcribe jobs can fail, of course, but users can't retry if in-progress marker is kept around.

### DIFF
--- a/misc/whisper_pod_transcriber/api.py
+++ b/misc/whisper_pod_transcriber/api.py
@@ -92,7 +92,6 @@ async def transcribe_job(podcast_id: str, episode_id: str):
 
 @web_app.get("/api/status/{call_id}")
 async def poll_status(call_id: str):
-    import modal.exception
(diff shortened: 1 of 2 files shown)


misc/whisper_pod_transcriber/api.py
```diff
@@ -103,7 +102,7 @@ async def poll_status(call_id: str):
         function_call.get(timeout=0.1)
     except TimeoutError:
         pass
-    except modal.exception.RemoteError as exc:
+    except Exception as exc:
```
thundergolfer, line 105: Hmm, so apparently you don't _always_ get a `RemoteError`: 

image

PR #94, opened 2022-11-08
Update example webhook URLs after 'example-' prefixing

thundergolfer:
cc @erikbern 

I will also delete the old apps from our org. (It's a minor bit of a deployment awkwardness that an app rename involves manual delete.)


erikbern:
Thanks!

Yeah I agree it's a bit annoying when you get stray apps from renaming

PR #98, opened 2022-11-14
whisper fixes: bad link and avoid race condition on podcast load

thundergolfer:
* Fix a misdirecting link on front page
* Remove unused import
* Avoid race condition that was causing errors like `FileNotFoundError: [Errno 2] No such file or directory: '/cache/podcast_metadata/649594/metadata.json'` to show up as 500s in the client.

### DIFF
--- a/misc/whisper_pod_transcriber/api.py
+++ b/misc/whisper_pod_transcriber/api.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 from typing import List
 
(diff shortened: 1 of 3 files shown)


thundergolfer:
Next issue to tackle is https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1668389848481249?thread_ts=1668322749.755669&cid=C031Z7H15DG. It's quite rare though. I've only encountered it in maybe 3 transcription attempts out of 100-200.

thundergolfer:
Deployed (v18) with `MODAL_ENV=modal_labs modal app deploy whisper_pod_transcriber.main`

PR #104, opened 2022-11-22
adds example for pet art generation with Dreambooth

charlesfrye:
This adds an example, based on the FastAI+W&B+Gradio CIFAR10 example, for finetuning a Stable Diffusion model with textual inversion.

It deploys a basic Gradio app. You can check out the final app [here](https://charlesfrye-example-dreambooth-app-fastapi-app.modal.run). It's associated with user `charlesfrye`'s app `example-dreambooth-app`.

Note: I wasn't able to run all of the pre-commit checks. Because of an issue with node versions and glibc, I couldn't run `prettier`. There is one css file, copied from the CIFAR10 example.

I also didn't see a way to build the docs locally, so I couldn't check the MD formatting etc.


thundergolfer:
🎊 looks great in-situ

image

Does the images file need accessed via a URL? I think it could be easier to have the user pass a filepath to a file that sits on their local disk, and that can just just read into a `list[str]`.

I tried running the training, and got this: 

```python
OSError: Can't load tokenizer for 'CompVis/stable-diffusion-v1-4'. If you were trying to load it from 'https://huggingface.co/models', make sure you
don't have a local directory with the same name. Otherwise, make sure 'CompVis/stable-diffusion-v1-4' is the correct path to a directory containing
all relevant files for a CLIPTokenizer tokenizer.
```

Have you seen that error before?

PS. Great looking Golden. Miss the days when my sister's was that small (which seems like only yesterday!)

misc/dreambooth_app.py
```diff
@@ -0,0 +1,331 @@
...
+            f"--instance_prompt='{prompt}'",
+            f"--resolution={config.resolution}",
+            f"--train_batch_size={config.train_batch_size}",
+            f"--gradient_accumulation_steps={config.gradient_accumulation_steps}",
+            f"--learning_rate={config.learning_rate}",
+            f"--lr_scheduler={config.lr_scheduler}",
+            f"--lr_warmup_steps={config.lr_warmup_steps}",
+            f"--max_train_steps={config.max_train_steps}",
+        ]
+    )
```
erikbern, line 238: might be good to add something like `stdout=sys.stdout, stderr=sys.stderr, check=True` to `subprocess.run` in case the process has any issues

misc/dreambooth_app.py
```diff
@@ -0,0 +1,331 @@
...
+        script_content = from_file.readlines()
+        with open("train_dreambooth.py", "w") as to_file:
+            to_file.writelines(script_content)
+
+    # define the training prompt
+    instance_phrase = f"{config.instance_name} {config.class_name}"
+    prompt = f"{config.prefix} {instance_phrase} {config.postfix}".strip()
+
+    # run training -- see huggingface accelerate docs for details
+    subprocess.run(
```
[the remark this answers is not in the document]
erikbern: I guess we could import it and run `parse_args([all the orgs])` and then pass that to `main`. That would at least obviate the need for the extra arguments I mentioned in the other comment (and exceptions would be properly handled etc). Not super important though.

erikbern:
I'm not sure why the checks aren't running. There should be an "approve and run" button, but I don't see it.

misc/dreambooth_app.py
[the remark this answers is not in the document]
charlesfrye: Agreed! I think that may have a bit to do with `accelerate`, which really wants to manipulate scripts -- it's used for multinode distributed training, which (at least in PyTorch) involves sending `.py` scripts around.

charlesfrye:
@thundergolfer:
> I tried running the training, and got this:
```python
OSError: Can't load tokenizer for 'CompVis/stable-diffusion-v1-4'. If you were trying to load it from 'https://huggingface.co/models', make sure you
don't have a local directory with the same name. Otherwise, make sure 'CompVis/stable-diffusion-v1-4' is the correct path to a directory containing
all relevant files for a CLIPTokenizer tokenizer.
```
> Have you seen that error before?

I've seen this error when the HF download fails. That's usually because the HF account hasn't accepted the license terms. They should be at the top of the page [here](https://huggingface.co/CompVis/stable-diffusion-v1-4).

If that's the cause of your error, then I'd welcome suggestions for improving the help text or making failure more informative. There is not a programmatic/CLI-based way to accept the LICENSE, but there is [an open issue about it](https://github.com/huggingface/huggingface_hub/issues/1198) where we could chime in in favor of the feature.

misc/dreambooth_app.py
charlesfrye: Wait, by default the subprocess inherits the stdout/stderr from the parent, and FWICT that works fine with `modal app run` here.

Also, noting here the  most of the work happens in `train_dreambooth.py`, which `accelerate` is also calling in a subprocess. It is called with `check=True`, so most errors are already handled. But adding `check` will catch issues at the level of the `accelerate` execution.

charlesfrye:
@thundergolfer:
> Does the images file need accessed via a URL? I think it could be easier to have the user pass a filepath to a file that sits on their local disk, and that can just just read into a list[str].

They do not. I've updated the script to work off of local files as of 43c8f332e32d8d409d536d2a32c8ca8d50ce922c.

I structured the code this way when I was still confused about how to pass information from local to Modal -- CLI args, local files, and local env. I was convinced I was supposed to use `modal app run X`, rather than `python X`, and I couldn't figure out how to do CLI args especially, so I just made everything available over the public internet.

If I understand correctly, the preferred pattern for using CLI args is as you described: inside of a `with stub.run`, pass them in to the function(s) being called. The pattern for local files is to use a `Mount`, and the pattern for local env variables is to use a `Secret` that's attached to the `Stub` via a key. Does that souynd right?

misc/dreambooth_app.py
erikbern: > Wait, by default the subprocess inherits the stdout/stderr from the parent

Oh nice, I didn't realize that

```
>>> subprocess.run(["echo", "xyz"])
xyz
```

erikbern (approved):
I think this is a great addition, so let's get it merged and we can always iterate more on it!

erikbern:
I had to disable the required checks for this to be possible to merge – not sure why Github wouldn't let me otherwise

aksh-at:
Thanks for contributing this example once again @charlesfrye! It's now live at https://modal.com/docs/guide/ex/dreambooth_app

PR #107, opened 2022-11-23
fix: expire in-progress transcription jobs

thundergolfer:
We use a `modal.Dict` to avoid running multiple transcription jobs for the same podcast and provide a nice user experience on hard-refresh of the browser. _But_, we never expire the `Dict` values, and in the worst case this causes the following error: 

image

The input ID is presumably so old that its been deleted from DB.


thundergolfer:
Added a couple of bonus fixes.

PR #108, opened 2022-11-24
New top-level 'data' section

thundergolfer:
Further structuring of the `misc/` section, which had become too crowded.

This section can now (and in future) contain:

* ELT and reverse ELT stuff (Meltano, Airbyte)
* DuckDB, Datasette
* Analytics stuff (Apache Arrow, Pandas)
* Data Warehouse stuff (Snowflake, DBT)
* Stats stuff


erikbern:
Would be good to prefix with a number so there's some sort of natural order? But idk

thundergolfer:
Yep I was wondering if the 1-10 ordering was fixed and `integrations/` and `misc/` were 'the rest'. 

Happy to fix this and number 11-14 the unnumbered folders.

erikbern:
I don't have any super strong feelings about it tbh. Maybe for symmetry we could do it

PR #111, opened 2022-11-29
Optimized Stable Diffusion 1.5 example with CLI

luiscape:
This adds an optimized Stable Diffusion 1.5 tutorial to our docs. The breakdown is:

- **A100**: 6.5s to load model and 1.6s per generation
- **T4**: 13s to load model and 3.7s per generation

One of the key changes is to use the pre-trained `diffusers.EulerAncestralDiscreteScheduler`. It is fast and generates high quality images with fewer steps (20-30). Other optimizations such as tracing are available [here](https://huggingface.co/docs/diffusers/optimization/fp16#memory-and-speed), so it may be possible to make the generation step faster.

### DIFF
--- a/06_gpu/stable_diffusion_cli.py
+++ b/06_gpu/stable_diffusion_cli.py
@@ -27,18 +27,27 @@ app = typer.Typer()
 #
 # Your model will be running remotely inside a container. We will be installing
 # all the model dependencies in the next step. We will also be "baking the model"
-# into the image using the script `download_stable_diffusion_models.py`. 
+# into the image using the script `download_stable_diffusion_models.py`.
 # This is technique that allows you to copy model files to
 # a worker more efficiently because they only need to be moved once.
 
-image = modal.Image.conda().apt_install(["curl"]).run_commands([
-    "conda install xformers -c xformers/label/dev",
-    "conda install pytorch torchvision pytorch-cuda=11.7 -c pytorch -c nvidia",
-]).run_commands([
-    "pip install diffusers[torch] transformers ftfy accelerate"
-]).run_commands([
-    "curl -L https://gist.github.com/luiscape/36a8cd29b8ed54cfbfcf56d51fe23cc0/raw/a6bf16996efe7c59114eea7944b0f99741d83d54/download_stable_diffusion_models.py | python"
-], secrets=[modal.Secret.from_name("huggingface-secret")])
+image = (
+    modal.Image.conda()
+    .apt_install(["curl"])
+    .run_commands(
+        [
+            "conda install xformers -c xformers/label/dev",
+            "conda install pytorch torchvision pytorch-cuda=11.7 -c pytorch -c nvidia",
+        ]
+    )
+    .run_commands(["pip install diffusers[torch] transformers ftfy accelerate"])
+    .run_commands(
+        [
+            "curl -L https://gist.github.com/luiscape/36a8cd29b8ed54cfbfcf56d51fe23cc0/raw/a6bf16996efe7c59114eea7944b0f99741d83d54/download_stable_diffusion_models.py | python"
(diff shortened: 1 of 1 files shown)


06_gpu/stable_diffusion_cli.py
```diff
@@ -0,0 +1,118 @@
...
+    .run_commands(
+        [
+            "conda install xformers -c xformers/label/dev",
+            "conda install pytorch torchvision pytorch-cuda=11.7 -c pytorch -c nvidia",
+        ]
+    )
+    .run_commands(["pip install diffusers[torch] transformers ftfy accelerate"])
+    .run_commands(
+        [
+            "curl -L https://gist.github.com/luiscape/36a8cd29b8ed54cfbfcf56d51fe23cc0/raw/a6bf16996efe7c59114eea7944b0f99741d83d54/download_stable_diffusion_models.py | python"
```
erikbern, line 46: This seems pretty hacky – do you think you can use the `run_function` method on the image? It lets you run a Python function during image construction, which I think is what you need here?

06_gpu/stable_diffusion_cli.py
```diff
@@ -0,0 +1,118 @@
...
+def entrypoint(prompt: str, samples: int = 10, steps: int = 20):
+    typer.echo(f"prompt => {prompt}, steps => {steps}, samples => {samples}")
+
+    dir = Path("/tmp/stable-diffusion")
+    if not dir.exists():
+        dir.mkdir(exist_ok=True, parents=True)
+
+    with stub.run():
+        for i in range(samples):
+            image = _run_inference(prompt, steps)
```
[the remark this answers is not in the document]
erikbern: (not important, just a thought)

06_gpu/stable_diffusion_cli.py
```diff
@@ -0,0 +1,118 @@
...
+# make subsequent calls to the same predict method much faster given that they
+# no longer need to instantiate the model. We'll get performance improvements
+# using this technique.
+#
+# We have also have applied a few model optimizations to make the model run
+# faster. On an A100, the model takes about 6.5s to load into memory, and then
+# 1.6s per generation on average. On a T4, it takes 13s to load and 3.7s per
+# generation. Other optimizations are also available [here](https://huggingface.co/docs/diffusers/optimization/fp16#memory-and-speed).
+
+if stub.is_inside():
```
erikbern, line 66: It might be nicer to use container lifecycle methods here.

06_gpu/stable_diffusion_cli.py
```diff
@@ -0,0 +1,118 @@
...
+# ## Model dependencies
+#
+# Your model will be running remotely inside a container. We will be installing
+# all the model dependencies in the next step. We will also be "baking the model"
+# into the image using the script `download_stable_diffusion_models.py`.
+# This is technique that allows you to copy model files to
+# a worker more efficiently because they only need to be moved once.
+
+image = (
+    modal.Image.conda()
```
erikbern, line 35: Not super important but is it worth doing this without conda? Since conda is kind of slow and harder to work with (imo)

luiscape: I had a version without conda but I am not able to compile xformers without it yet.

Inference is about 20% slower without xformers installed.

erikbern (approved):
In the interest of getting this out, I think we should merge it asap and keep iterating on it. I'm happy to try some of the things I mentioned in my review – I can create a PR later today.

@luiscape feel free to merge whenever

06_gpu/stable_diffusion_cli.py
luiscape: Yes, that's a good idea. It's better than piping that script.

PR #112, opened 2022-11-29
Misc updates to the fast stable diffusion

erikbern:
The main changes:
* Instead of downloading a script and running it (during the image build phase), inline that script into the code itself
* Use container lifecycle hooks to load the model from disk

### DIFF
--- a/06_gpu/stable_diffusion_cli.py
+++ b/06_gpu/stable_diffusion_cli.py
@@ -46,19 +46,14 @@ def download_models():
     # Download the Euler A scheduler configuration. Faster than the default and
     # high quality output with fewer steps.
     euler = diffusers.EulerAncestralDiscreteScheduler.from_pretrained(
-        model_id,
-        subfolder="scheduler",
-        use_auth_token=hugging_face_token,
-        cache_dir=cache_path)
+        model_id, subfolder="scheduler", use_auth_token=hugging_face_token, cache_dir=cache_path
+    )
     euler.save_pretrained(cache_path)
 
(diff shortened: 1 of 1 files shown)


06_gpu/stable_diffusion_cli.py
```diff
@@ -41,11 +67,10 @@
...
     )
     .run_commands(["pip install diffusers[torch] transformers ftfy accelerate"])
-    .run_commands(
-        [
-            "curl -L https://gist.github.com/luiscape/36a8cd29b8ed54cfbfcf56d51fe23cc0/raw/a6bf16996efe7c59114eea7944b0f99741d83d54/download_stable_diffusion_models.py | python"
-        ],
+    .run_function(
+        download_models,
         secrets=[modal.Secret.from_name("huggingface-secret")],
+        gpu=modal.gpu.A100(),
```
luiscape, line 73: I believe that this step doesn't need to run on a GPU.

luiscape (approved):

06_gpu/stable_diffusion_cli.py
erikbern: It doesn't have to, but there's a minor warning when you don't.

```
/usr/local/lib/python3.9/site-packages/diffusers/models/attention.py:433: UserWarning: Could not enable memory efficient attention. Make sure xformers is installed correctly and a GPU is available: torch.cuda.is_available() should be True but is False. xformers' memory efficient attention is only available for GPU
```

PR #113, opened 2022-11-30
Replace old meltano example with updated UI hosting version

freider:
The old meltano example wasn't working and wasn't updated to use modal shared volumes etc. for persisting state.

This new one takes a slightly new approach:
* Meltano project is locally managed using the meltano cli
* Upon (Modal) runs, an image is built using the meltano.yml file, where all dependencies are installed etc. and a shared volume is set up to store the meltano database and run logs.
* Running `python meltano_ui.py` (or `modal app deploy meltano_ui.py`) now deploys a wsgi app with the (deprecated) meltano UI interface.

Ideas for improvements:
* Custom Modal `orchestrator` for meltano, which would deploy all meltano jobs and schedules as Modal functions w/ the specified run interval. The Modal UI would then basically replace the old meltano UI, allowing log inspection of individual runs and kicking of specific jobs on demand.


data/meltano/meltano_project/meltano.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: staging
+- name: prod
+plugins:
+  extractors:
+  - name: tap-github
+    variant: meltanolabs
+    pip_url: git+https://github.com/MeltanoLabs/tap-github.git
+    config:
+      repositories:
+      - modal-labs/synchronicity
```
thundergolfer, line 15: What does this do?

thundergolfer (commented):
Nice 👍 

What is running when I go into the UI and click run now? It doesn't seem like the Modal app does the compute. 

When I click run now I got a pipeline failure because of a HTTP 401 in Github API call

<kbd>
image
</kbd>

data/meltano/meltano_project/meltano.yml
freider: It configures a github "tap" (data extractor) for fetching commit info from the synchronicity repo (any public repo would work). It then uses a `jsonl` "loader" (data exporter) on that data.

freider:
> What is running when I go into the UI and click run now? It doesn't seem like the Modal app does the compute.
It should be running within the Modal webhook (it's just the built in wsgi app hosted on Modal), but now that I think about it I wonder what happens if the web request gets cancelled etc. Will have to dig into the meltano ui source code to find out if it spawns a thread or something...  In that case it might get cancelled once the container shuts down due to inactivity of inputs. Using the meltano UI might be a bad idea for other reasons too though :thinking:  
 
> When I click run now I got a pipeline failure because of a HTTP 401 in Github API call
Did you set up the required github secret? It requires an auth token to use the github graphql api I think

thundergolfer:
> Did you set up the required github secret?

No I didn't. I didn't realize that I had to.

freider:
Modal should probably print a big warning when a user deploys an app referencing non-existing secrets - I thought we did...

thundergolfer:
We throw an error. I ran the app under `modal_labs` account, so the secret should have existed?

freider:
> We throw an error. I ran the app under `modal_labs` account, so the secret should have existed?

Ah so apparently I had set the secret on my personal account when testing it, but we also had a secret with the same name (but a different env var name...) on modal_labs. Updated the modal_labs secret now so it should work

PR #115, opened 2022-11-30
removed superfluous arguments to from_pretrained

erikbern:
* `cache_path` is provided twice, which seems unnecessary
* The floating point precision is already determined by the weights (running some benchmarks and removing this argument actually might even make it _faster_, although I'm not 100% sure)


erikbern:
Merging this for now, happy to address any comments though.

PR #123, opened 2022-12-01
Use safetensors

erikbern:
This doesn't lead to any speedup.

### DIFF
--- a/06_gpu/stable_diffusion_cli.py
+++ b/06_gpu/stable_diffusion_cli.py
@@ -51,13 +51,13 @@ def download_models():
     euler = diffusers.EulerAncestralDiscreteScheduler.from_pretrained(
         model_id, subfolder="scheduler", use_auth_token=hugging_face_token, cache_dir=cache_path
     )
-    euler.save_pretrained(cache_path)
+    euler.save_pretrained(cache_path, safe_serialization=True)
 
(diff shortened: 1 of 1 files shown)


06_gpu/stable_diffusion_cli.py
```diff
@@ -107,7 +116,7 @@ def __enter__(self):
         self.pipe = diffusers.StableDiffusionPipeline.from_pretrained(cache_path, scheduler=euler).to("cuda")
         self.pipe.enable_xformers_memory_efficient_attention()
 
-    @stub.function(gpu=modal.gpu.A100())
+    @stub.function(gpu=modal.gpu.A100(), secret=modal.Secret({"SAFETENSORS_FAST_GPU": "1"}))
```
luiscape, line 119: `SAFETENSORS_FAST_GPU` needs to be enabled when we load the models in `__enter__`. When adding it here, would it be available there?

luiscape (commented):
See comment

06_gpu/stable_diffusion_cli.py
erikbern: yeah it will be available. this is a confusing thing about how the lifecycle methods work – and why i want to move the decorator to be on the class instead

PR #132, opened 2022-12-06
mypy check script

thundergolfer:
Trying to pass the right combination of arguments and flags to get `mypy` working in a multi-project repo was too annoying, so knitting everything together in a Python script.

So far type-checking is only done on the 01-10 folders. I can't yet enable type-checking on the packages in misc, ml, and data because there are too many type issues. But can progressively fix those issues and then introduce `py.typed` marker files to opt-in packages to type-hint checking. 

**aside:** Why do we prefix the scripts with `example_`? Could we instead move scripts into a `scripts/` folder and remove the prefix?

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -51,6 +51,23 @@ jobs:
       - name: Check isort
         run: isort . -c -v --profile black
 
+  mypy:
+    name: MyPy
+    runs-on: ubuntu-20.04
+
+    steps:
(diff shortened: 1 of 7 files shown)


erikbern:
> **aside:** Why do we prefix the scripts with `example_`? Could we instead move scripts into a `scripts/` folder and remove the prefix?

I think they existed elsewhere at some point. Feel free to move them!

PR #134, opened 2022-12-07
fix: modal.aio_lookup isnt used in ex, and doesn't even exist

thundergolfer:
Some old lint I just noticed when working on another, related doc.


erikbern:
It should be `modal.aio.aio_lookup` – https://github.com/modal-labs/modal-client/blob/main/modal/aio.py

but this works too

erikbern:
Oh cool, yeah that's a bit problematic if it's gone

PR #136, opened 2022-12-08
Update 1-10 examples code for github.com/modal-labs/modal-client/pull/97

thundergolfer:
I also addressed some nits that I came across when running these examples just now. 

Next up is the larger examples in `data/` and `ml/`.

### DIFF
--- a/03_scaling_out/youtube_face_detection.py
+++ b/03_scaling_out/youtube_face_detection.py
@@ -101,7 +101,7 @@ def detect_faces(fn, start, stop):
 
 # ### Modal entrypoint function
 #
-# The "entrypoint" into Modal controls the main flow of the program:
+# This 'entrypoint' into Modal controls the main flow of the program:
 #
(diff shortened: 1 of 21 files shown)


03_scaling_out/youtube_face_detection.py
```diff
@@ -119,15 +119,15 @@ def run(url):
...
-    out_fns = list(detect_faces.starmap(args))
+    print("Processing each range of 1s intervals using a Modal map")
+    out_fns = list(detect_faces.starmap(intervals))
 
-    # Convert to video clips
+    print("Converting detections to video clips")
     out_clips = [moviepy.editor.VideoFileClip(out_fn) for out_fn in out_fns]
 
-    # Concatenate results
+    print("Concatenating results")
```
thundergolfer, line 130: Without this printing, this example can hang with no console output for ~10-15 seconds, which is not so nice.

PR #138, opened 2022-12-08
fix: updated datasette ver requires initialization step

thundergolfer:
Noticed this was broken just now when I went to test it. Consequences of not pinning package versions in our image definitions...


thundergolfer:
Already deployed it to fix the linked app in the blog post.

PR #139, opened 2022-12-09
Use io.BytesIO context manager in duckdb example

vegarsti:
Use io.BytesIO's context manager to close the buffer after using it.


erikbern:
Nice! I think we do something similar in a few other examples so we can fix that too

vegarsti:
> Nice! I think we do something similar in a few other examples. I can fix it. Thanks!

Great!

ekzhang:
Thank you @vegarsti!

PR #141, opened 2022-12-09
Revert "Temp fix for Stable Diffusion Slackbot docs"

thundergolfer:
Reverts modal-labs/modal-examples#129

I can't reproduce this 'missing paragraph' problem locally. Using `-` or `*` bullets both work. Possibly fixed by going back to 1 column layout, which happened 1 day after Akshat noticed the issue?  cc @aksh-at


thundergolfer:
If this re-introduces the problem on modal.com, then something strange is going on and I'll look more into it.

PR #154, opened 2022-12-19
Updated meltano example using .copy() instead of hacks

freider:


10_integrations/meltano/meltano_modal.py
```diff
@@ -1,55 +1,45 @@
...
+    subprocess.check_call(["meltano", "install"])
 
 
 meltano_img = (
     modal.Image.debian_slim()
     .apt_install(["git"])
     .pip_install(["meltano"])
-    .extend(dockerfile_commands=invalidating_on_update)
-    .run_function(meltano_install, mounts=[meltano_source_mount])
+    .copy(meltano_source_mount, "/meltano")
```
freider, line 36: A bit ugly since mounting stuff at the default path ("/") isn't possible at the moment, pending a worker PR

erikbern (approved):

PR #159, opened 2022-12-20
Capitalize A10G correctly

ekzhang:
https://www.nvidia.com/en-gb/data-center/products/a10-gpu/

### DIFF
--- a/06_gpu_and_ml/dreambooth_app.py
+++ b/06_gpu_and_ml/dreambooth_app.py
@@ -264,7 +264,7 @@ def train(instance_example_urls, config=TrainConfig()):
 
 @stub.asgi(
     image=image,
-    gpu="A10g",  # Don't need as much VRAM for inference, so can use a cheaper GPU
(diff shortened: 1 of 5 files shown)


jkukul:
@ekzhang Could it be that this actually broke something? 

The example dreambooth app seems to be broken:
https://modal-labs-example-dreambooth-app-fastapi-app.modal.run/

I can reproduce the error when deploying the app myself, but when I change it back to `gpu="A10g"` it works.

erikbern:
Thanks for mentioning @jkukul  – we'll take a look!

jkukul:
After more testing, I think this PR  unrelated. The app breaks for me with both `A10G` and `A10g`.  In both cases, the contents of the volume are not attached to `/model` directory. 

It consistently works with `A100`, so the same gpu that was used for training the model.

erikbern:
It sounds like what you're running into @jkukul is that A100s run in a different cloud provider than A10G, so the shared volume content will be different based on that. It's pretty confusing and we hope to fix this soon!

PR #190, opened 2023-01-19
Fix stable diffusion

erikbern:
1. Use https://github.com/modal-labs/modal-client/pull/221 to install the latest xformers prerelase
2. Fix the `stub.run` inside `@local_entrypoint` (which actually was _working_, but i think added some extra latency)

### DIFF
--- a/06_gpu_and_ml/stable_diffusion_cli.py
+++ b/06_gpu_and_ml/stable_diffusion_cli.py
@@ -91,8 +91,8 @@ image = (
         "transformers",
         "triton",
         "safetensors",
-        "xformers==0.0.16rc393",
     )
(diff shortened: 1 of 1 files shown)


06_gpu_and_ml/stable_diffusion_cli.py
```diff
@@ -170,18 +170,17 @@ def entrypoint(prompt: str, samples: int = 5, steps: int = 10, batch_size: int =
     if not dir.exists():
         dir.mkdir(exist_ok=True, parents=True)
 
-    with stub.run():
```
freider, line 173: Oops, my bad, missed this when fixing the examples. Interesting that it still worked 😬

PR #191, opened 2023-01-19
Fix dreambooth container image

erikbern:
* Install latest `xformers` using `pip install --pre`
* Install the same version of diffusers as the train_dreambooth.py script

### DIFF
--- a/06_gpu_and_ml/dreambooth_app.py
+++ b/06_gpu_and_ml/dreambooth_app.py
@@ -97,7 +97,7 @@ class TrainConfig(SharedConfig):
     postfix: str = ""
 
     # locator for plaintext file with urls for images of target instance
-    instance_example_urls_file: str = "dreambooth_app/instance_example_urls.txt"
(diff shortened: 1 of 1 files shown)


aksh-at (approved):
Cool, nice that consistency between the train_dreambooth script and diffusers version is enforced!

06_gpu_and_ml/dreambooth_app.py
```diff
@@ -42,20 +42,26 @@
...
+GIT_SHA = "ed616bd8a8740927770eebe017aedb6204c6105f"
+
+image = (
+    modal.Image.debian_slim()
+    .pip_install(
+        "accelerate",
+        "datasets",
+        "ftfy",
+        "gradio~=3.10",
+        "smart_open",
```
aksh-at, line 54: Do we still need this?

erikbern: it's still used for downloading images, but i'll change that in a subsequent PR

PR #199, opened 2023-01-23
use modal serve with whisper demo

erikbern:
Cleaning up a bunch of entrypoints in the examples

### DIFF
--- a/06_gpu_and_ml/whisper_pod_transcriber/main.py
+++ b/06_gpu_and_ml/whisper_pod_transcriber/main.py
@@ -6,7 +6,6 @@ import dataclasses
 import datetime
 import json
 import pathlib
-import sys
 from typing import Iterator, Tuple
 
 import modal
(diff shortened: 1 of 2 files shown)


06_gpu_and_ml/whisper_pod_transcriber/main.py
```diff
@@ -411,16 +410,9 @@ def fetch_episodes(show_name: str, podcast_id: str, max_episodes=100):
...
-        with stub.run():
-            index()
-    elif cmd == "search-podcast":
-        with stub.run():
-            for pod in search_podcast.call(sys.argv[2]):
-                print(pod)
-    else:
-        exit(f"Unknown command {cmd}. Supported commands: [transcribe, run, serve, index, search-podcast]")
+@stub.local_entrypoint
+def search_entrypoint(name):
```
thundergolfer, line 414:
```suggestion
def search_entrypoint(name: str):
```

Would work anyway, but good habit to have the types as they determine arg parsing.

thundergolfer (approved):

06_gpu_and_ml/whisper_pod_transcriber/main.py
erikbern: yeah good point, thanks!

PR #202, opened 2023-01-23
vision model: change entrypoint

erikbern:


06_gpu_and_ml/vision_model_training.py
```diff
@@ -290,20 +292,24 @@ def fastapi_app():
...
+# To run training as an ephemeral app:
+#
+# ```shell
+# modal run vision_model_training.py::stub.train
+# ```
+#
+# To test the model on an image, run:
+#
+# ```shell
+# modal run vision_model_training.py::stub.classify_url --image-url <url>
```
thundergolfer, line 306: Do we have any ideas about how to make these ref less complex?

erikbern: @freider yeah maybe we could make it a bit more user friendly?

PR #209, opened 2023-01-27
black formatter line length reduction

thundergolfer:
**motivation:**  80 chars fits on modal.com/docs code blocks. 120 does not. We could adjust the UI to be higher, but we're unlikely to go all the way to 120.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ pythonpath = [
 ]
 
 [tool.black]
-line-length = 120
+line-length = 80
 
 [tool.mypy]
 ignore_missing_imports = true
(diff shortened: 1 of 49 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ pythonpath = [
 ]
 
 [tool.black]
-line-length = 120
+line-length = 80
```
erikbern, line 12: could probably remove this section – i assume this is the default anyway

erikbern:
i generally kinda hate the default width of 80, but it actually makes sense for docs

LGTM!

pyproject.toml
thundergolfer: >  Black defaults to 88 characters per line,

erikbern: that's so weird

PR #217, opened 2023-02-06
Adds example of setting up a tailscale sidecar

freider:
For accessing private networks via a userspace proxy

### DIFF
--- a/10_integrations/modal_tailscale.py
+++ b/10_integrations/modal_tailscale.py
@@ -1,0 +1,92 @@
+"""Example of setting up tailscale VPN (https://tailscale.com/) as a sidecar
+
+In this example, it's used to access a private API in another machine on the tailnet
+The `tailscale_sidecar` utility function contains most of the setup, and needs to
+be called from the container defined in `tailscale_image`.
+"""
+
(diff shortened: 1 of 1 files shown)


10_integrations/modal_tailscale.py
```diff
@@ -0,0 +1,92 @@
...
+        f"--authkey={tailscale_authkey}",
+        "--hostname=modal-app",
+    ]
+    working_tailscale = threading.Event()
+
+    def output_watcher(tailscaled_output):
+        for line in tailscaled_output:
+            if show_output:
+                print(f"TAILSCALE: {line}", file=sys.stderr)
+            if b"magicsock: derp-1 connected" in line:
```
freider, line 59: Seems like the proxy/network isn't working until this line appears, which doesn't happen until a short while *after* `tailscale up` finishes

freider, line 59: Reading a bit about it, it seems DERP is not used all the time with tailscale, so this would not always work, and sometimes cause issues. Perhaps safer to just add a sleep(5) or something :thinking:

PR #230, opened 2023-02-23
provide a Conda-less Tensorflow GPU option in Tensorflow ex

thundergolfer:
A TreeHacks hackathon team wanted to use Tensorflow and so grabbed this example code. Then, they unsurprisingly got stuck with some Conda problem when they tried to extend the image. 

Here I add a non-Conda 'Tensorflow with GPUs' image option.

### DIFF
--- a/06_gpu_and_ml/tensorflow_tutorial.py
+++ b/06_gpu_and_ml/tensorflow_tutorial.py
@@ -13,26 +13,32 @@
 # ## Setting up the dependencies
 #
 # Installing Tensorflow in Modal is quite straightforward.
-# If you want it to run on a GPU, it's easiest to use the base Conda image.
-# We also need to install `cudatoolkit` and `cudnn` for it to work.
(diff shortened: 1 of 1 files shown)


06_gpu_and_ml/tensorflow_tutorial.py
```diff
@@ -13,26 +13,32 @@
...
-        modal.Image.conda()
-        .conda_install(
-            "cudatoolkit=11.2",
-            "cudnn=8.1.0",
-            "cuda-nvcc",
-            channels=["conda-forge", "nvidia"],
-        )
-        .pip_install("tensorflow~=2.9.1")
-    ),
+    image=conda_image or dockerhub_image,  # pick one and remove the other.
```
luiscape, line 41: Why not leave only the non-conda one?

thundergolfer: Because I think this is a good test of `conda` to have around. It runs in synmon.

PR #259, opened 2023-03-29
basic Alpaca LoRA example

thundergolfer:
<kbd>
image
</kbd>

<br>-

**TODO**

Will leave the following as follow-up changes:

- [ ] Remove warning from bitsandbytes: ` UserWarning: The installed version of bitsandbytes was compiled without GPU support. 8-bit optimizers and GPU quantization are unavailable.`. I think it's erroneous, as the code is explicitly installing the CUDA version of the package.
- [ ] Reduce cold-start by storing the base pre-trained model in the Modal image with [`run_function`](https://modal.com/docs/reference/modal.Image#run_function).


erikbern:
Super cool! How bad is the cold start?

PR #266, opened 2023-04-04
fix deploy of text-to-pokemon

thundergolfer:
Addressing this: 

image


06_gpu_and_ml/text-to-pokemon/text_to_pokemon/main.py
```diff
@@ -174,7 +174,11 @@ def diskcached_text_to_pokemon(prompt: str) -> list[bytes]:
 
 
 @stub.asgi(
```
erikbern, line 176: this needs to be updated to the newest web endpoint syntax – weird that the tests didn't cathc that

thundergolfer: Will do.

PR #275, opened 2023-04-11
make 'hello world' ex clear about local/remote execution

thundergolfer:
Motivated by https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1681242971024609 I went looking for a place in our docs to add more about the `f()` vs `f.call()` distinction. 

> Aside: The log streaming was definitely more sluggish than it used to be. It used to be so snappy. 

cc @erikbern as this is a prominent example and you may have thoughts.


erikbern (commented):
Nice! Starting to think maybe we should use .remote just like Ray btw

PR #276, opened 2023-04-13
fix: LangChain example code rot

thundergolfer:
This example started failing a couple of days ago and I wasn't checking syn-mon carefully enough to notice it.  A user pointed it out in beta-testing: https://modalbetatesters.slack.com/archives/C031Z7H15DG/p1681380968649189. 

Deployed v3 https://modal.com/apps/modal-labs/example-langchain-qanda.


06_gpu_and_ml/langchains/potus_speech_qanda.py
```diff
@@ -98,43 +98,10 @@ def retrieve_sources(sources_refs: str, texts: list[str]) -> list[str]:
     ]
 
 
-def create_retrying_openai_embedder():
```
thundergolfer, line 101: This is now handled internally by LangChain.

PR #277, opened 2023-04-17
whisper fine tune mar4

thundergolfer:
**TODO:**

- [ ] Setup conf/creds for pushing the training results to the Huggingface benchmark leaderboard


06_gpu_and_ml/openai_whisper/finetuning/train/transcribe.py
```diff
@@ -0,0 +1,101 @@
+import os
+import pathlib
+import sys
+from typing import TYPE_CHECKING
+
+from .config import app_config
+from .logs import get_logger
+
+import modal
+from modal.cli.volume import FileType
```
thundergolfer, line 10: Not ideal.

Issue #282, opened 2023-04-19
Dreambooth example - Unexpected behavior when calling a stub function directly


kr4t0n:
# Versions
- OS: ubuntu
- modal-client: 0.48.1946

# Intro
When playing around with [dreambooth example](https://modal.com/docs/guide/ex/dreambooth_app), the documentation said `you can kick off a training job with the command modal run dreambooth_app.py::stub.train`. When doing so, you may encounter an error as `AttributeError: 'str' object has no attribute 'xxx'`.

# Reproduction
The reproduction code is simple, just have a snippet as

```python
import modal

from dataclasses import dataclass

stub = modal.Stub('dataclass-test')


@dataclass
class SharedConfig:
    a: str = 'a'


@stub.function()
def test(config=SharedConfig()):
    print(f'Type: {type(config)}, Content: {config}')


@stub.local_entrypoint()
def main():
    test.call()
```

when running command `modal run test.py`, you may have normal output as 

```shell
ubuntu:~/projects/modal_test$ modal run test.py 
✓ Initialized. View app at xxx
✓ Created objects.
├── 🔨 Created test.
├── 🔨 Created mount /home/ubuntu/projects/modal_test/test.py
└── 🔨 Created fix.
Type: <class 'test.SharedConfig'>, Content: SharedConfig(a='a')
✓ App completed.
```

when kicking off the stub function directly `modal run test.py::stub.test`, you may notice that `config` became a string.

```shell
ubuntu:~/projects/modal_test$ modal run test.py::stub.test
✓ Initialized. View app at xxx
✓ Created objects.
├── 🔨 Created test.
├── 🔨 Created mount /home/ubuntu/projects/modal_test/test.py
└── 🔨 Created fix.
Type: <class 'str'>, Content: SharedConfig(a='a')
✓ App completed.
```

From my point of view, I think this behavior is unexpected.

# Workaround
A fix workaround is rather simple, just initialize the `config` inside the function rather sent as a parameter.

```python
@stub.function()
def fix():
    config = SharedConfig()
    print(f'Type: {type(config)}, Content: {config}')
```

Now, running stub function `modal run test.py::stub.fix` will behave correctly.

```shell
ubuntu:~/projects/modal_test$ modal run test.py::stub.fix
✓ Initialized. View app at xxx
✓ Created objects.
├── 🔨 Created test.
├── 🔨 Created mount /home/ubuntu/projects/modal_test/test.py
└── 🔨 Created fix.
[...]

PR #283, opened 2023-04-19
Dreambooth - Config as a string when kicking off the stub function directly

Kr4t0n:
A simple workaround to avoid this, detailed information could be found in Issue #282.


luiscape (approved):
LGTM

Thank you for the fix!

PR #284, opened 2023-04-19
initial demo of stable lm

stanbiryukov:
Initial demo of running zero-shot text generation with Stability AI's [stablelm-tuned-alpha-7b](https://huggingface.co/stabilityai/stablelm-tuned-alpha-7b). Converted model checkpoints to `safetensors` format and used `accelerate` to manage device mapping for faster container load times.

Inspired by Modal's [alpaca_lora](https://github.com/modal-labs/modal-examples/blob/e994cd6f8411232f49b9485aa8fdb0fc3114047c/06_gpu_and_ml/alpaca/alpaca_lora.py) example.

### DIFF
--- a/06_gpu_and_ml/stable_lm/main.py
+++ b/06_gpu_and_ml/stable_lm/main.py
@@ -20,7 +20,6 @@ def build_models():
     model_path = snapshot_download(
         "stabilityai/stablelm-tuned-alpha-7b",
         ignore_patterns=["*.md"],
-        revision="25071b093c15c0d1cb2b2876c6deb621b764fcf5",
     )
     m = AutoModelForCausalLM.from_pretrained(
         model_path,
@@ -42,11 +41,13 @@ image = (
     .conda_install(
         "cudatoolkit-dev=11.7",
         "pytorch-cuda=11.7",
-        channels=["nvidia", "pytorch"],
+        "rust=1.69.0",
+        channels=["nvidia", "pytorch", "conda-forge"],
     )
     .env(
         {
             "HF_HOME": "/root",
+            "HF_HUB_ENABLE_HF_TRANSFER": "1",
             "SAFETENSORS_FAST_GPU": "1",
             "BITSANDBYTES_NOWELCOME": "1",
             "PIP_DISABLE_PIP_VERSION_CHECK": "1",
@@ -98,6 +99,8 @@ class CompletionResponse(BaseModel):
 class StabilityLM:
     def __init__(self, model_url: str = "stabilityai/stablelm-tuned-alpha-7b"):
         self.model_url = model_url
+        os.environ["HF_HUB_OFFLINE"] = "1"
+        os.environ["TRANSFORMERS_OFFLINE"] = "1"
 
     def __enter__(self):
         """
--- a/06_gpu_and_ml/stable_lm/main.py
+++ b/06_gpu_and_ml/stable_lm/main.py
@@ -2,6 +2,9 @@ import base64
 import os
 import time
 from pathlib import Path
+from typing import List, Optional, Union
+
+from pydantic import BaseModel
 
 import modal
 
@@ -13,9 +16,21 @@ def build_models():
     import torch
     from huggingface_hub import snapshot_download
     from transformers import AutoModelForCausalLM, AutoTokenizer
-    model_path = snapshot_download("stabilityai/stablelm-tuned-alpha-7b", ignore_patterns=["*.md"], revision="25071b093c15c0d1cb2b2876c6deb621b764fcf5")
-    m = AutoModelForCausalLM.from_pretrained(model_path, torch_dtype=torch.float16, device_map="auto", local_files_only=True)
-    m.save_pretrained(model_path, safe_serialization=True, max_shard_size="24GB")
+
+    model_path = snapshot_download(
+        "stabilityai/stablelm-tuned-alpha-7b",
+        ignore_patterns=["*.md"],
+        revision="25071b093c15c0d1cb2b2876c6deb621b764fcf5",
+    )
+    m = AutoModelForCausalLM.from_pretrained(
+        model_path,
+        torch_dtype=torch.float16,
+        device_map="auto",
+        local_files_only=True,
+    )
+    m.save_pretrained(
+        model_path, safe_serialization=True, max_shard_size="24GB"
+    )
     tok = AutoTokenizer.from_pretrained(model_path)
     tok.save_pretrained(model_path)
     [p.unlink() for p in Path(model_path).rglob("*.bin")]  # type: ignore
@@ -29,11 +44,19 @@ image = (
         "pytorch-cuda=11.7",
         channels=["nvidia", "pytorch"],
     )
-    .env({"HF_HOME": "/root", "SAFETENSORS_FAST_GPU": "1", "BITSANDBYTES_NOWELCOME": "1", "PIP_DISABLE_PIP_VERSION_CHECK": "1", "PIP_NO_CACHE_DIR": "1"})
+    .env(
+        {
+            "HF_HOME": "/root",
+            "SAFETENSORS_FAST_GPU": "1",
+            "BITSANDBYTES_NOWELCOME": "1",
+            "PIP_DISABLE_PIP_VERSION_CHECK": "1",
+            "PIP_NO_CACHE_DIR": "1",
+        }
+    )
     .run_commands(
         f"echo '{requirements}' | base64 --decode > /root/requirements.txt",
         "pip install -r /root/requirements.txt",
-        gpu="A10G"
+        gpu="A10G",
     )
     .run_function(
         build_models,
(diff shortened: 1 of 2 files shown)


06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,146 @@
...
+from pydantic import BaseModel, Field
+from rich import print
+from transformers import (
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    GenerationConfig,
+    StoppingCriteria,
+    StoppingCriteriaList,
+    pipeline,
+)
```
thundergolfer, line 15:
```suggestion
from transformers import (
    StoppingCriteria,
    StoppingCriteriaList,
    pipeline,
)
```

Ideally we don't need to import `transformers` in global scope, because it isn't used locally.

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,146 @@
+import time
+from pathlib import Path
+
+import torch
+from huggingface_hub import snapshot_download
```
thundergolfer, line 5: Can inline this import inside `compile_models` so it's not needed locally.

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,146 @@
...
+    .run_function(
+        fetch_models,
+        gpu=None,
+        timeout=3600,
+    )
+    .run_function(
+        compile_models,
+        gpu="A10G",
+        timeout=3600,
+    )
```
thundergolfer, line 57:
```suggestion
    .run_function(
        compile_models,
        gpu="A10G",
        timeout=3600,
    )
```

I think here you're trying to avoid using a GPU on the step which is just network-bound, but splitting into two functions makes the build less efficient. What's happening under the hood is that the first function runs in a new container the resulting filesystem state is baked as an image layer. Then the second function starts a new container with the previous image layer as a base. But in order to use the previously built image layer as a base it needs to download it. So this is still downloading the snapshot twice: once from Huggingface CDN and once from within Modal.

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,146 @@
...
+    return StabilityLM().generate.call(
+        prompt, temperature=temperature, max_new_tokens=max_new_tokens
+    )
+
+
+@stub.local_entrypoint()
+def main():
+    q_style, q_end = "[bold bright_magenta]", "[/bold bright_magenta]"
+    instructions = [
+        "Generate a list of 20 great names for sentient cheesecakes that teach SQL.",
```
thundergolfer, line 142:
```suggestion
        "Generate a list of the 10 most beautiful cities in the world..",
```

StableLM does very poorly on this cheesecake prompt compared to GPT-4 😄. 

```
1. Liam
2. Mila
3. Jadine
4. Pike
5. Olive
...
```

Made another suggestion, which is easier and where the answer seems more reasonable, though I'd hardly count NYC as the 3rd most _beautiful_ city in the world.

thundergolfer (commented):
Thanks heaps for putting this up! It's great to get example contributions from the community. 

I've left some specific comments on some things, but generally it'd be good to minimize the amount of third-party packages imported in global scope because they require the user to install them locally. 

For any third-party packages you don't want to use locally, e.g. `rich`, could you add them into a `requirements.txt` list? 🙏

stanbiryukov:
Thanks for the review and guidance, @thundergolfer! I'll have some time over the weekend to polish this up and incorporate feedback. My goal with the example is to get it to a point that demonstrates how one could host an open-source llm on Modal that mimics OpenAI's create completions spec for use as a drop-in replacement. Aiming to build on the whisper example for handling the concurrency and thread lock logic for the streaming component.

thundergolfer:
Sounds great 👌

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,146 @@
...
+        gpu="A10G",
+        timeout=3600,
+    )
+)
+
+stub = modal.Stub(name="example-stability-lm", image=image)
+
+
+class StopOnTokens(StoppingCriteria):
+    def __call__(self, input_ids: torch.LongTensor, scores: torch.FloatTensor, **kwargs) -> bool:
```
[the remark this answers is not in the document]
stanbiryukov: Great point and thanks for the `TYPE_CHECKING` reference. It might also be best to add within the model class to handle various stop words.

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: Agreed. I just didn't want to re-run the image build at the first pass.

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: All makes sense, thanks for the explanation! We can combine these.

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: Agreed on minimizing local imports

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,141 @@
...
+
+def compile_models():
+    from transformers import AutoModelForCausalLM, AutoTokenizer
+    MODEL_PATH = snapshot_download("stabilityai/stablelm-tuned-alpha-7b", ignore_patterns=["*.md"], revision="25071b093c15c0d1cb2b2876c6deb621b764fcf5")
+    m = AutoModelForCausalLM.from_pretrained(MODEL_PATH, torch_dtype=torch.float16, local_files_only=True).cuda()
+    tok = AutoTokenizer.from_pretrained(MODEL_PATH, local_files_only=True)
+    m = torch.compile(m)
+    m.save_pretrained(MODEL_PATH, safe_serialization=True, max_shard_size="24GB")
+    tok.save_pretrained(MODEL_PATH)
+    [p.unlink() for p in Path(MODEL_PATH).rglob("*.bin")]
```
thundergolfer, line 32:
```suggestion
    [p.unlink() for p in Path(MODEL_PATH).rglob("*.bin")]  # type: ignore
```

Seeing `06_gpu_and_ml/stable_lm/run.py:34: error: "unlink" of "Path" does not return a value` from mypy

stanbiryukov: Thanks for the heads up.

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,322 @@
...
+    .conda_install(
+        "cudatoolkit-dev=11.7",
+        "pytorch-cuda=11.7",
+        "rust=1.69.0",
+        channels=["nvidia", "pytorch", "conda-forge"],
+    )
+    .env(
+        {
+            "HF_HOME": "/root",
+            "HF_HUB_ENABLE_HF_TRANSFER": "1",
```
thundergolfer, line 51: Nice. Did you observe a speed up from this? If so, how much?

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,322 @@
...
+        build_models,
+        gpu=None,
+        timeout=3600,
+    )
+)
+
+stub = modal.Stub(
+    name="example-stability-lm",
+    image=image,
+    secrets=[modal.Secret({"REPO_ID": "stabilityai/stablelm-tuned-alpha-7b"})],
```
thundergolfer, line 73: What's this env var used for? I see that this ID string appears in a number of other places, so wonder if this one is redundant.

thundergolfer (approved):
Great stuff. Thanks again for putting this up 💯

06_gpu_and_ml/stable_lm/main.py
```diff
@@ -0,0 +1,322 @@
...
+    [p.unlink() for p in Path(model_path).rglob("*.bin")]  # type: ignore
+
+
+image = (
+    modal.Image.conda()
+    .apt_install("git", "software-properties-common", "wget")
+    .conda_install(
+        "cudatoolkit-dev=11.7",
+        "pytorch-cuda=11.7",
+        "rust=1.69.0",
```
thundergolfer, line 45: Is this for `hf-transfer`? That package is written in Rust but shouldn't need the toolchain installed because it has published wheels on PyPi: https://pypi.org/project/hf-transfer/#files. Or is `conda` ignoring those PyPi wheels...

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: I'd have to quantify. During build, I didn't notice much of a change for this file size but locally on an M1 I do find it quicker for fetching larger model files.

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: Good point it's redundant. I was trying to solve the original issue where I was re-writing this in a few places by putting in main stub but didn't end up using :) Ideally we can abstract a bit more to handle passing in various models and sizes as they are released.

06_gpu_and_ml/stable_lm/main.py
stanbiryukov: Good point we're not building from source since it's compiled in the wheel. I didn't try without it though and conda does respect the `pip` installs we have so removing this would make the image leaner.

PR #286, opened 2023-04-21
fix: batch inf example should use new lifecyle API

thundergolfer:
It's failing CI over in https://github.com/modal-labs/modal-examples/pull/285


erikbern:
Thanks!

PR #294, opened 2023-04-26
Youtube example reliability, and some maintenance

thundergolfer:
Following up on the Youtube example having ~50% reliability in the past few days. Seems PyTube has started being unreliable. There's about 5 issues about `KeyError: 'streamingData'` which is the error we're intermittently getting.


erikbern:
Nice!

PR #298, opened 2023-05-01
Fine-tuning LLM for Spam Detection

thundergolfer:
A fine-tuning example that has these new features compared with existing examples:

* The training dataset is prepped and stored in a Modal persistent volume (`dataset.py`)
* Multiple, increasingly powerful model architectures are defined and are trainable (`models.py` and `train.py`)
* Model tracking and storage using a Modal persistent volume  (`model_registry.py` and `model_storage.py`)
* Model deployment from persistent volume storage based on model ID (`serving.py`)
* Model serving that connects inference to served model metadata (`serving.py`)

I think this is so far the most fully developed example of doing MLOps on Modal, but more to be done around:

* Multi-model serving
* Trained model evaluation and comparison


erikbern:
Nice, this looks like a lot of work. Would love a walkthrough at some point!

thundergolfer:
Happy to demo at some point.

PR #327, opened 2023-05-16
DBT duckDB example, and fixes to DB SQLite example

thundergolfer:
* Inspired by [Ted Conbeer's code snippet](https://modalbetatesters.slack.com/archives/C031WA6KLGM/p1682699742110879) in the beta Slack I've made a DBT x DuckDB example.
* Pinned the deps in the SQLite DBT example as the latest versions don't work. 
* Renames the existing DBT SQLite example to fit better with the new example

**TODO:**

- [ ] This example becomes synmon testable if we add the relevant secret in the `modal_labs` account. I tested this using my personal account. Would have to create a new AWS user with permissions to do CRUD on the example's S3 bucket.


tconbeer (approved):
This looks great -- you might consider renaming your `seed` function to `create_source_data`, so you don't confuse people with `dbt seed`, but otherwise I think this is a good example!

PR #328, opened 2023-05-16
fix: example reliability and configuration fixes sweep

thundergolfer:
Most notably deletes `pymc_pmf` as this example had been failing and left unmaintained for months.


misc/webscraper.py
```diff
@@ -31,7 +31,7 @@ async def get_links(url: str):
         )
         await browser.close()
 
-    return links
+    return set(links)
```
thundergolfer, line 34: We're hitting rate limits from Slack posting so many links. This will reduce the many duplicates we attempt to post, and hopefully alleviate the rate-limiting issues.

PR #339, opened 2023-06-05
text-to-pokemon: Gracefully handle missing paths

ksu22:
### Type of Change

- [ ] New Example

## Checklist

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/text-to-pokemon/README.md
+++ b/06_gpu_and_ml/text-to-pokemon/README.md
@@ -13,6 +13,7 @@ excellent CSS work.
 
 ```bash
 cd "$(git rev-parse --show-toplevel)/06_gpu_and_ml/text-to-pokemon/text_to_pokemon/frontend"
+npm install
 npx vite build --watch
 ```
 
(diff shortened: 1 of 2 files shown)


06_gpu_and_ml/text-to-pokemon/text_to_pokemon/main.py
```diff
@@ -112,11 +60,15 @@ def image_to_byte_array(image) -> bytes:
         return buf.getvalue()
 
 
+@stub.cls(gpu="A10G", shared_volumes={config.CACHE_DIR: volume}, keep_warm=1)
 class Model:
     def __enter__(self):
-        self.pipe = load_stable_diffusion_pokemon_model().to("cuda")
+        import threading
+        if not pokemon_naming.rnn_names_output_path.exists():
```
ksu22, line 67: not thread safe at scale but most of these examples aren't either, going to just defer to precedent

thundergolfer (approved):
Thanks for the fixes and improvements! It's great that running the name gen as a manual step isn't required anymore.

PR #341, opened 2023-06-07
Run black to fix linter

ksu22:
### Type of Change

- [ ] New Example

## Checklist

## Outside contributors


ksu22:
forgot that github actions arent enabled by default when forking so i missed the `black` linter check, so definitely wanted to fix that. figured id be a good samaritan by also fixing `PyTest`

erikbern:
Nice, thanks for doing this!!

PR #347, opened 2023-06-20
stable diffusion: torch 2.0.1+cu117

danielnorberg:
Update stable diffusion example to use `torch` `2.x` as required by latest `xformers`.

### Type of Change

- [ ] New Example


luiscape (approved):
👍

PR #350, opened 2023-06-25
MPT-30B example (8-bit with bitsandbytes)

gongy:


06_gpu_and_ml/mpt_bitsandbytes.py
```diff
@@ -0,0 +1,145 @@
...
+def cli():
+    question = "Can you describe the main differences between Python and JavaScript programming languages."
+    model = MPT30B()
+    for text in model.generate.call(prompt_template.format(question)):
+        print(text, end="", flush=True)
+
+
+@stub.function(timeout=60 * 10)
+@web_endpoint()
+def get(question: str):
```
ekzhang, line 133: This function can just run in the base debian_slim image right?

PR #361, opened 2023-07-16
pytube fix is no longer needed

rchalamala:
…o main.

### Type of Change

- [ ] New Example

## Checklist

## Outside contributors


thundergolfer (approved):
Thanks! 🙏

PR #382, opened 2023-07-30
ocr: fix modal.Image#run_function ref path

danielnorberg:
Fix broken path seemingly introduced in https://github.com/modal-labs/modal-examples/pull/381


thundergolfer (commented):
Thanks

PR #384, opened 2023-07-30
Enable ruff isort rules on the repo

aksh-at:
Most of the diff is from prettier on files that weren't previously linted


aksh-at:
Going to merge this in since it's not disruptive to any of the in-flight PRs

ekzhang:
thanks for running prettier too 🌟

PR #388, opened 2023-08-02
fix: disable interactivity in Lambda

thundergolfer:


10_integrations/dbt/dbt_sqlite.py
```diff
@@ -80,8 +80,12 @@ def debug():
     dbt_cli.call(["debug"])
 
 
+# Only used in testing to disable interactivity.
+interactivity_enabled = os.environ.get("AWS_LAMBDA_RUNTIME_API") is None
```
ekzhang, line 84: Uhh not sure if we want to add this to the actual example, I would probably just remove it :')

Another thing we could do is not register interactive= functions unless `sys.stdout.isatty` or something? Or catch the tcgetattr error.

ekzhang (approved):

PR #391, opened 2023-08-04
Example for hosting vLLM as an API server

gongy:


gongy:
Superseded by `vllm_mixtral`

PR #393, opened 2023-08-04
Remove unnecessary token from Stable Diffusion CLI

ekzhang:


luiscape:
It was the case that we needed to accept terms of use before being able to download the model hence the need for the token. Great it's not required anymore!

PR #400, opened 2023-08-10
sdxl torch compilation comment

kousun12:
nice work putting up an sdxl example. i was just reading through and noticed this comment - updating it to be slightly more descriptive

### Type of Change

- [ ] New Example

## Checklist

## Outside contributors


aksh-at:
Just saw this, thanks for the fix!

PR #440, opened 2023-09-05
Use add_python in examples

ekzhang:


ekzhang:
The mlc_inference example is failing when I try to rebuild the image, due to an MLC wheel packaging snafu yesterday https://github.com/mlc-ai/package/issues/53

PR #443, opened 2023-09-07
Use stub instead of container_app in podcast transcriber

aksh-at:


ekzhang (approved):
Thanks for the explanation!

thundergolfer (approved):

PR #458, opened 2023-10-03
Add Img2Img Example for Stable Diffusion

michaeleliot:
Adds an example for img2img transformation for stable diffusion, taking in both a query and image. This combines some aspects of the dreambooth, and the xl examples. Would love feedback, especially with comments.

I'll also add the xl example seems out of date with it's dependencies. Happy to update if that's of interest.

Screenshot 2023-10-03 at 1 12 39 AM

### Type of Change

- [x] New Example

## Checklist

## Outside contributors


kning:
this example i think is well enough covered in this tutorial: https://modal.com/docs/examples/stable_diffusion_xl_turbo#stable-diffusion-xl-turbo-image-to-image, so will close this for now. thanks for the suggestion though!

PR #473, opened 2023-10-20
Add A1111 example with tunnels (bad)

ekzhang:
Web launcher that starts a new A1111 web UI at a dedicated URL, then sends it to the requester.

This is fine, but it takes a while to start up. Most of the example is figuring out how to install A1111 correctly. :(

image

### Type of Change

- [x] New Example

## Checklist


06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
```diff
@@ -0,0 +1,139 @@
...
+        "libgl1",
+        "libglib2.0-0",
+        "google-perftools",  # For tcmalloc
+    )
+    .env({"LD_PRELOAD": "/usr/lib/x86_64-linux-gnu/libtcmalloc.so.4"})
+    .run_commands(
+        "git clone --depth 1 --branch v1.6.0 https://github.com/AUTOMATIC1111/stable-diffusion-webui /webui",
+        "python -m venv /webui/venv",
+        "cd /webui && . venv/bin/activate && "
+        + "python -c 'from modules import launch_utils; launch_utils.prepare_environment()' --xformers",
```
erikbern, line 39: could probably use `run_function` for this?

06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
```diff
@@ -0,0 +1,139 @@
...
+    --mixed_precision=fp16 \
+    --dynamo_backend=inductor \
+    --num_cpu_threads_per_process=6 \
+    /webui/launch.py \
+        --skip-prepare-environment \
+        --listen \
+        --port 8000
+"""
+    with forward(8000) as tunnel:
+        p = subprocess.Popen(START_COMMAND, shell=True)
```
erikbern, line 67: i guess there's no way to import it and launch it programmatically?

06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
```diff
@@ -0,0 +1,139 @@
...
+            headers={"Cache-Control": "public, max-age=0, must-revalidate"},
+        )
+
+    @app.post("/launch")
+    def launch(password: str = Form(), timeout: int = Form()):
+        if password != app_password:
+            raise HTTPException(401, "Incorrect password")
+
+        key = str(uuid.uuid4())  # for Dict polling
+        start_web_ui.spawn(key, timeout)
```
erikbern, line 132: i think this is a case where it would be nice to have the ability for the _caller_ to set up the tunnel rather than the container itself – something like `.spawn_with_tunnel(...)` and it returns an url

06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
ekzhang: I think the run_function would have to enter the venv again though :/

06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
ekzhang: Yeah, there's no way to do this. Also requires `accelerate`

06_gpu_and_ml/stable_diffusion/a1111_tunnel.py
ekzhang: Agreed, although it would already not require the extra Dict/Queue if FunctionCall.get() worked with generators.

PR #485, opened 2023-11-01
Add Migration from Celery queue to Modal example

yirenlu92:
Guide that goes over migrating from using a celery queue for async video processing to Modal


kning:
@yirenlu92 still interested in merging this?

yirenlu92:
@kning yes I think it's ready to merge

yirenlu92:
oh I don't think I have push permissions to this repo anymore...

09_job_queues/video_processing.py
```diff
@@ -0,0 +1,191 @@
...
+# ## The Modal application
+
+# To use Modal instead of a Celery queue, we first create a separate file, `modal_queue.py`, and define a Modal Stub and Image.
+# The image is a Debian image with ffmpeg, wget, git, and the python dependencies installed.
+
+# We also import the helper function from `tasks.py`, `transcribe_video_and_extract_screenshots`, which is doing the bulk of the video processing work.
+# Note that we are importing `transcribe_video_extract_screenshots` from outside of the Modal function. This is because the `tasks.py` file exists locally.
+
+import modal
+from tasks import transcribe_video_and_extract_screenshots
```
kning, line 105: pytest is complaining about this import, how do you want to handle? tried breaking out tasks you have commented out above into a separate file, but that one also seems to refer to some other modules i can't seem to find.

yirenlu92: Yeah, I think this example was previously supposed to be on the blog, where the code snippets don't actually have to run...

I guess one option is that I can turn `transcribe_video_and_extract_screenshots` into a dummy function, commenting out its body, and just make note of that in the comments above. Issue is that part of what I wanted to show-case was the migration from a non-Modal job queue to a Modal job queue. I also wanted to show how to import different modules.

Are there no other multi-file examples?

charlesfrye:
discussed closure offline

PR #488, opened 2023-11-03
Add more fine-tuning result images to the Diffusers tutorial

yirenlu92:
- more examples? ideally like 20 instead of 3
- resize images
- include what the prompts were for those images
- put pics at the top
- deploy gradio app and link to it

https://www.loom.com/share/85fa5f31334d4e288c70d225d89275fd


06_gpu_and_ml/diffusers/train_and_serve_diffusers_script.py
```diff
@@ -11,10 +11,119 @@
...
+# ```
+
+# ```
+# file_name,text
+# arrow.png,"In the HCON style, an icon of an arrow"
+# bike.png,"In the HCON style, an icon of an arrow"
+# cruiseShip.png,"In the HCON style, an icon of an arrow"
+# ```
+
+# 4. Upload the dataset to HuggingFace Hub.
```
erikbern, line 103: should we just put the dataset on a volume instead? i think it's better the more we can keep things within our own ecosystem

erikbern (approved):
nice, the images look great!

06_gpu_and_ml/diffusers/train_and_serve_diffusers_script.py
yirenlu92: I think it's going to be harder for people to run the example out of the box if they have to download all the heroicon images/process them themselves. I just included the processing steps so they know how do it if they want to do it with their own dataset.

yirenlu92: oh, do you mean we spin up a `PersistentVolume` ourselves?

erikbern: honestly not sure – will think about it

PR #489, opened 2023-11-06
Fix mlc_inference.py error in model_lib_path, upgrade to CUDA 12.2.2

rgreer4:
Updates this example from ChatModule(model, lib_path) which will currently fail to run, to ChatModule(model, model_lib_path).  

The CUDA 12.1.0 image is deprecated/scheduled for deletion, so also ups to current 12.2.2.

### Type of Change

- [ ] New Example
- [X] Example updates (Bug fixes, new features, etc.)


luiscape:
Thank you. Appreciate the update.

PR #490, opened 2023-11-07
Fix examples now that we test against Python 3.11

thundergolfer:
Recently we changed our example testing system to use Python 3.11. In future we should test examples for compatibility against multiple Python versions, but for now I've just added a disclaimer to the README.

### Type of Change

- [ ] New Example


ekzhang:
Thank you for working through these!

PR #491, opened 2023-11-07
Streamlit example

thundergolfer:
<kbd>
image
</kbd>

### Type of Change

- [x] New Example

## Checklist

## Outside contributors

### DIFF
--- a/10_integrations/streamlit/serve_app.py
+++ b/10_integrations/streamlit/serve_app.py
@@ -43,26 +43,34 @@ streamlit_script_remote_path = pathlib.Path("/root/app.py")
 #
 # We could also import the module, and then pass `app.__path__` to Streamlit.
 
+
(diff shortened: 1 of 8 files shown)


10_integrations/streamlit/serve_app.py
```diff
@@ -0,0 +1,92 @@
...
+# This technique is very similar to what is shown in the [Tunnels guide](/docs/guide/tunnels).
+
+
+@stub.function()
+@modal.web_endpoint(method="GET")
+def share():
+    from fastapi.responses import RedirectResponse
+
+    run_streamlit.spawn(publish_url=True)
+    url = stub.q.get()
```
erikbern, line 91: we've talked about it before but for things like this it would be amazing to have something like `.spawn_relay(...)` that returns an url

PR #496, opened 2023-11-13
Remove isort from CI

ekzhang:
This is made redundant by Ruff.

Also updating CI to use Python 3.11, which is now standard here.


erikbern:
That reminds me, I think Ruff also replaces Black now: https://docs.astral.sh/ruff/formatter/#black-compatibility

ekzhang:
Wow that's neat!

PR #501, opened 2023-11-21
Basic code for SVD on streamlit

aksh-at:


06_gpu_and_ml/stable_video_diffusion.py
```diff
@@ -0,0 +1,85 @@
...
+            flag_options={},
+        )
+
+
+@stub.function()
+@modal.web_endpoint(method="GET", label="svd")
+def share():
+    from fastapi.responses import RedirectResponse
+
+    run_streamlit.spawn(publish_url=True)
```
erikbern, line 83: would be amazing to have a `.spawn_with_tunnel` method so we don't have to use a queue for this

PR #503, opened 2023-11-27
fix: typo

pingren:
### Type of Change

- [ ] New Example

## Checklist

## Outside contributors


pingren:
I didn't know https://peps.python.org/pep-0515/ Thanks for clarification!

PR #506, opened 2023-12-01
Update Streamlit example

luiscape:
Updates Streamlit example to:

1. pin dependencies: avoids breaking example if dependencies update
2. installs missing dependency: `streamlit` now requires `git` to be installed
3. adds links to scripts: made it clearer that we need to create two scripts in order to run example

### DIFF
--- a/10_integrations/streamlit/serve_streamlit.py
+++ b/10_integrations/streamlit/serve_streamlit.py
@@ -12,8 +12,8 @@
 #
 # The example is structured as two files:
 #
-# 1. This module, which defines the Modal objects.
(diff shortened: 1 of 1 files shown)


10_integrations/streamlit/serve_streamlit.py
```diff
@@ -23,7 +23,11 @@
 # The `app.py` script imports three third-party packages, so we include these in the example's
 # image definition.
 
-image = modal.Image.debian_slim().pip_install("streamlit", "numpy", "pandas")
+image = (
+    modal.Image.debian_slim()
+        .apt_install("git")
```
aksh-at, line 28: Is this needed for something?

PR #507, opened 2023-12-01
Example simplification using `@build`

erikbern:
Using https://github.com/modal-labs/modal-client/pull/1136 for some examples.

This is just one or two lines less for most examples, but for `stable_diffusion_cli` it's a drastic reduction since we can use the same function for both `@build()` and `@enter()`

### DIFF
--- a/06_gpu_and_ml/stable_diffusion/stable_diffusion_cli.py
+++ b/06_gpu_and_ml/stable_diffusion/stable_diffusion_cli.py
@@ -56,29 +56,6 @@ model_id = "runwayml/stable-diffusion-v1-5"
 cache_path = "/vol/cache"
 
 
-def download_models():
-    import diffusers
-    import torch
-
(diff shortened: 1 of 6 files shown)


06_gpu_and_ml/stable_diffusion/stable_diffusion_cli.py
```diff
@@ -118,14 +98,28 @@ def download_models():
...
 
-@stub.cls(gpu="A10G")
+@stub.cls(image=image, gpu="A10G")
 class StableDiffusion:
-    def __enter__(self):
-        import diffusers
-        import torch
-
-        torch.backends.cuda.matmul.allow_tf32 = True
+    def _download_models(self):
```
erikbern, line 103: some of this method might be redundant with `__enter__` potentially

thundergolfer (approved):

PR #510, opened 2023-12-01
Add example for Image2Image SDXL turbo

RohanArepally:
CleanShot 2023-12-01 at 16 51 34@2x

CleanShot 2023-12-01 at 16 51 16@2x


aksh-at (approved):
Awesome, looks great!

PR #513, opened 2023-12-05
Scaffolding for `gpt-fast` example

irfansharif:
Show how to run https://github.com/pytorch-labs/gpt-fast on Modal. It's incomplete.

### Checklist


aksh-at (commented):
Instead of essentially vendoring the gpt-fast repo, is it possible to clone the original `gpt-fast` as part of the image build, and only add the code here that's necessary for running it within Modal?

I'm not sure how feasible that is, but it's hard to tell at a glance which code is new vs existing here.

kning:
hey @irfansharif still working on this? or should we close and revisit?

irfansharif:
I'll abandon it for now.

PR #535, opened 2024-01-03
Fix failing synthetic monitors

ekzhang:
We've had synthetic monitors failing for several weeks now.

Going through all of them to see what the problem is.


10_integrations/algolia_indexer.py
```diff
@@ -73,7 +73,8 @@
 
 
 @stub.function(
-    image=algolia_image, secrets=[Secret.from_name("algolia-secret")]
+    image=algolia_image,
+    secrets=[Secret.from_name("algolia-secret", environment_name="main")],
```
erikbern, line 77: i think this might be an issue with the synthetic monitor code maybe? we should probably use the "main" environment for all examples

ekzhang: Hm it's coded to be the `syn-mon-{runtime}-{cloud}` environment right now

ekzhang: I thought about it a bit, don't see any harm in changing to `main`

ekzhang:
Okay I think I looked through and fixed all of modal-examples, except the text_embeddings_inference and tgi_mixtral

10_integrations/algolia_indexer.py
erikbern: ok i remember now! i think another option is to keep synthetic monitors in a separate environment, but in that case we need to move all secrets to that env which is a bit annoying

seems fine to move examples back to main

erikbern (approved):
Noticed you deleted the DBT example – is it broken beyond repair? No strong feelings about it

ekzhang:
I kept the `dbt_duckdb` example, is that alright? The `dbt_sqlite` example depends on the `meltano_modal` example, and that one isn't running at the moment.

erikbern:
Sure seems fine! Should we delete the Meltano example too then?

ekzhang:
Okay making a separate PR to delete meltano!

thundergolfer (commented):
LGTM. Appreciate you getting onto this!

PR #543, opened 2024-01-06
ComfyUI example

thundergolfer:
image

Works pretty well! I quite like UI, reminds me a lot of [Grasshopper](https://www.backofficepro.com/engineering/grasshopper-in-architecture.php) from my architecture days.

### Type of Change

- [x] New Example

## Checklist


06_gpu_and_ml/stable_diffusion/comfy_ui.py
```diff
@@ -0,0 +1,148 @@
...
+    spawn_comfyui_in_background()
+
+    config = type(
+        "Config",
+        (BaseURLProxyConfigMixin, ProxyConfig),
+        {
+            "upstream_base_url": f"http://{HOST}:{PORT}",
+            "rewrite_host_header": f"{HOST}:{PORT}",
+        },
+    )()
```
ekzhang, line 147: hmm I think we should try to avoid repeating this `type()` magic

thundergolfer: Yeh I just copy-pasted this from the Streamlit example. I don't have the background on why it was setup this way.

hsiutayu:
When I download another model, Comfyui says null. Only the model included in the example works. 

Also, how do I persist the downloaded model in a modal volume?

PR #549, opened 2024-01-16
Create multion_news_agent.py

Div99:
### Type of Change

- [X ] New Example

## Checklist

- [X ] Example is testable in synthetic monitoring system, or `lambda-test: false` is added to example frontmatter

## Outside contributors

### DIFF
--- a/10_integrations/multion_news_agent.py
+++ b/10_integrations/multion_news_agent.py
@@ -1,0 +1,73 @@
+# ---
+# lambda-test: false
+# ---
+# # Twitter News Agent
+
+# In this example, we use Modal to deploy a cron job that periodically checks for cool AI news everyday and tweets it on Twitter using the MultiOn Agent API.
+
+# ## Import and define the stub
+#
+# Let's start off with imports, and defining a Modal stub.
+
+import os
+from datetime import datetime, timedelta
+
+import modal
+
+stub = modal.Stub("ai-news-tweet-agent")
+
+# ## Searching for AI News
+#
+
+QUERY = "serverless"
+WINDOW_SIZE_DAYS = 1
+
+# Let's also define an image that has the `requests` package installed, so we can query the API.
+
+multion_image = modal.Image.debian_slim().pip_install("multion")
+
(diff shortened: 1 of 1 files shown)


Div99:
@aksh-at

10_integrations/multion_news_agent.py
```diff
@@ -0,0 +1,73 @@
...
+from datetime import datetime, timedelta
+
+import modal
+
+stub = modal.Stub("ai-news-tweet-agent")
+
+# ## Searching for AI News
+#
+
+QUERY = "serverless"
```
aksh-at, line 22: I don't think these constants are used anywhere

10_integrations/multion_news_agent.py
```diff
@@ -0,0 +1,73 @@
...
+WINDOW_SIZE_DAYS = 1
+
+# Let's also define an image that has the `requests` package installed, so we can query the API.
```
aksh-at, line 25:
```suggestion
# Let's also define an image that has the `muliton` package installed, so we can query the API.
```

10_integrations/multion_news_agent.py
```diff
@@ -0,0 +1,73 @@
...
+
+multion_image = modal.Image.debian_slim().pip_install("multion")
+
+# We can now define our main entrypoint, that use MultiOn to scrape AI news everyday and post it on my twitter account. We specify a [schedule](/docs/guide/cron) in the function decorator, which
```
aksh-at, line 29:
```suggestion
# We can now define our main entrypoint, that uses [MultiOn](https://www.multion.ai/) to scrape AI news everyday and post it on our twitter account. We specify a [schedule](/docs/guide/cron) in the function decorator, which
```

aksh-at, line 29: Can we add a section that explains what MultiOn is, and how to set it up for this example? I.e. I have to make an account, install a chrome extension, log in to my Twitter (?)

aksh-at (changes requested):
great, thanks for adding this! Couple comments but main one is that it would be great to explain to users what MultiOn is and how to set it up :)

aksh-at (approved):
Looks like there's a couple typos, but good to go otherwise!

PR #552, opened 2024-01-16
init

jxnl:
This is a part of the code for the upcoming blog post. 

### Type of Change

- [x] New Example

## Checklist

@aksh-at can you explain this test thing? this job runs for a long time.

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/embeddings/wikipedia/README.md
+++ b/06_gpu_and_ml/embeddings/wikipedia/README.md
@@ -1,0 +1,97 @@
+# Embedding Wikipedia in 15 minutes
+
+This example shows how we can embed the entirety of english wikipedia on Modal in just 15 minutes. We've published a detailed writeup which walks you through the implemenation [here](#todo).
+
+## Description
+
+There are a total of 2 files in this repository
+
+- `download.py` : This showcases how to download the Wikipedia dataset into a `Modal` volume. We can take advantage of `Modal`'s high internet speeds to download large datasets quickly.
+
+- `main.py`: This showcases how to run an embedding job on your downloaded dataset and run a parallelizable job using Modal's inbuilt parallelization abstraction.
+
+## Getting Started
+
+You'll need a few packages to get started - we recommend using a virtual environment to install all of the dependencies listed in the `requirements.txt`
+
+```bash
+python3 -m venv venv
+source venv/bin/activate
+pip3 install -r requirements
+```
+
+Once you've done so, you'll need to authenticate with Modal. To do so, run the command `modal token new`.
+
+This will open up a new tab in your default browser and allow you to run, deploy and configure all of your Modal applications from your terminal.
+
+## Downloading Our Dataset
+
+Let's first download our Wikipedia dataset into a Modal volume. We can optimise the download time using the `num_proc ` keyword to parallelize some of the downloads.
+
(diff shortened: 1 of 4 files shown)


06_gpu_and_ml/embeddings/wikipedia/README.md
```diff
@@ -0,0 +1,97 @@
...
+- `main.py`: This showcases how to run an embedding job on your downloaded dataset and run a parallelizable job using Modal's inbuilt parallelization abstraction.
+
+## Getting Started
+
+You'll need a few packages to get started - we recommend using a virtual environment to install all of the dependencies listed in the `requirements.txt`
+
+```bash
+python3 -m venv venv
+source venv/bin/activate
+pip3 install -r requirements
```
gongy, line 20: Why is `datasets` and `httpx` necessary?

Could be nice to just have a `pip3 install modal`.

gongy: (Also if we can't remove the requirements entirely, might need to be `requirements.txt`)

06_gpu_and_ml/embeddings/wikipedia/main.py
```diff
@@ -0,0 +1,327 @@
...
+    "--port",
+    "8000",
+    "--max-client-batch-size",
+    str(BATCH_SIZE),
+    "--max-batch-tokens",
+    str(BATCH_SIZE * 512),
+]
+
+
+stub = Stub("embeddings")
```
gongy, line 45: I think we name something like `example-embeddings` in our examples so far.

06_gpu_and_ml/embeddings/wikipedia/main.py
```diff
@@ -0,0 +1,327 @@
+import asyncio
+import json
+import subprocess
+from modal import Image, Secret, Stub, Volume, gpu, method
+
+# We first set out configuration variables for our script.
+## Embedding Containers Configuration
+N_GPU = 100
```
gongy, line 8: What does `N_GPU` refer to? The concurrent requests per GPU?

06_gpu_and_ml/embeddings/wikipedia/main.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+@stub.cls(
+    gpu=GPU_CONFIG,
+    image=tei_image,
+    concurrency_limit=N_GPU,
+    retries=3,
+)
+class TextEmbeddingsInference:
+    def __enter__(self):
```
gongy, line 128: It could be quite nice to add a comment here that the Modal class syntax allows you to define a complex container lifecycle in one fell swoop.

The `__enter__` function runs once per GPU startup, and the `@method()` function runs once per inference on an initialized GPU.

06_gpu_and_ml/embeddings/wikipedia/main.py
gongy, line 8: Maybe `GPU_CONCURRENCY`?

06_gpu_and_ml/embeddings/wikipedia/main.py
gongy: Also, is there a reason we left out `allow_concurrent_inputs` here?

PR #582, opened 2024-02-13
Updates Tailscale example

luiscape:
Updates the Tailscale example to use a custom entrypoint instead of a sidecar. This adds all containers to the user's tailnet automatically, making them addressable via task ID.


aksh-at (commented):
Awesome!

PR #586, opened 2024-02-20
Convert examples to all use lifecycle method decorators

mwaskom:
We are deprecating the dunder hooks for lifecycle management (`__enter__`, `__exit__`, etc.) in favor of function decorators (`@enter`, `@exit`, etc.). See [the docs](https://modal.com/docs/guide/lifecycle-functions) for more details. This PR updates all deprecated code in the examples.

### Type of Change

- [ ] New Example

### DIFF
--- a/06_gpu_and_ml/embeddings/text_embeddings_inference.py
+++ b/06_gpu_and_ml/embeddings/text_embeddings_inference.py
@@ -99,14 +99,13 @@ class TextEmbeddingsInference:
         self.client = AsyncClient(base_url="http://127.0.0.1:8000")
 
     @exit()
-    def teardown_server(self):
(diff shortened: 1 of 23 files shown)


06_gpu_and_ml/embeddings/text_embeddings_inference.py
```diff
@@ -93,18 +93,19 @@ def download_model():
     allow_concurrent_inputs=10,
 )
 class TextEmbeddingsInference:
-    def __enter__(self):
+    @enter()
```
erikbern, line 96: we could probably also add `@build` here and skip the `run_function` step

erikbern (approved):
for a lot of these we should also consider using `@build` instead of `run_function` – doesn't have to be this PR though

PR #588, opened 2024-02-21
SDXL lightning example

aksh-at:


erikbern:
Can we collapse this to one function call instead of two?

PR #601, opened 2024-02-27
Fix path for llm-frontend reference in tgi.py

yirenlu92:
After refactor, llm-frontend and llm-serving are in the same parent directory


thundergolfer:
@yirenlu92 can this merge?

PR #602, opened 2024-02-27
add first draft comfy api example

kning:
Basically modal app version of the comfyUI [websockets example](https://github.com/comfyanonymous/ComfyUI/blob/8daedc5bf2ac106f1920c634866198c82e06997e/script_examples/websockets_api_example.py#L4).

This passes a comfy workflow json to a running comfyUI server, queues a job, and fetches the image outputs of that job via API. 

Chatted with @thundergolfer on how this could be expanded into a more standalone example:
- Put a simple frontend (like [this example](https://www.comfydeploy.com/share/comfy-deploy-example-txt2img-sdxl)) that allows a user to adjust the editable parts of a comfyui json workflow (e.g. text prompts).
- Use this code as a starting point web endpoint to handle those requests and create the workflow json, spin up a "headless" comfyui server, and return the output

Other thoughts:
* [fal.ai](https://fal.ai/models/comfy-executor/api) allows you to pass urls of models
* how would people adapt this if their workflow included plugins?

### DIFF
--- a/06_gpu_and_ml/stable_diffusion/comfy_api.py
+++ b/06_gpu_and_ml/stable_diffusion/comfy_api.py
@@ -1,0 +1,262 @@
+import modal
+import uuid
+
+stub = modal.Stub(name="example-comfy-api")
+image = modal.Image.debian_slim(python_version="3.10").pip_install("websocket-client")
+
+with image.imports():
+    import json
+    import urllib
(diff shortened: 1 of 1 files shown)


thundergolfer:
Superseded by https://github.com/modal-labs/modal-examples/pull/628

PR #612, opened 2024-03-01
moves slow mypy type checking out of main check script

charlesfrye:
I love a good type system, but I'm not convinced of the value of the mypy type check.

Most of our examples break due to rotting from insufficiently-controlled dependencies, rather than anything detectable with types. Recall that the types of functions on Modal are often not even checkable locally. Furthermore, type hints in other Python libraries in the ML ecosystem are often poorly instrumented and so the checker is subject to false positives and negatives. And finally, we have other checks in place to ensure examples run end-to-end, so the most common user facing errors can be caught directly.

It is also about 4x slower than the other checks we run in CI,  >2min instead of ~30s, which IME pushes "waiting on a PR" into context switch territory.

But rather than fully removing the check, this PR makes it a daily cronjob that is also triggerable manually. Whoever last changed the schedule (me, for now) will be notified of failures.

### Type of Change

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,23 +25,6 @@ jobs:
 
       - run: ruff format --check
 
-  mypy:
-    name: MyPy
-    runs-on: ubuntu-20.04
-
-    steps:
-      - uses: actions/checkout@v3
-
-      - uses: actions/setup-python@v4
-        with:
-          python-version: "3.11"
-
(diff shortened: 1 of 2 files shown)


ekzhang:
I personally think this is fine. But let's wait on a review before making devops changes like this in the future.

We spend a lot of time on our ops system and I would rather get feedback before changing it

PR #624, opened 2024-03-05
removes SD ONNX example

charlesfrye:
I think we should remove this example.

- I wasn't able to create an environment in which this example works, because the setup is pretty finicky (e.g. hardware aware, official guides say to build from source),
- ONNX doesn't seem to be particularly popular, and
- both [results from other teams](https://huggingface.co/blog/sdxl_ort_inference) and the original experiments when the example was introduced suggest the performance gains are minimal. `torch.compile` is pretty good!

### Type of Change


luiscape (approved):
Can you also remove the link from [this page](https://modal.com/docs/examples/misc-ml-examples)?

PR #634, opened 2024-03-06
Adds `CloudBucketMount` example mounting S3 bucket 

luiscape:
Adds example that shows how to use a `CloudBucketMount` in Modal. This example both writes and reads from an S3 mount. Both writes and reads are done in parallel using both `map` and `starmap`.

The example is similar to: https://modal.com/docs/examples/duckdb_nyc_taxi


charlesfrye (approved):
Made some changes, as discussed offline.

This is ready to merge once it passes in synmon -- aka once are all workers ready to run it.

PR #635, opened 2024-03-06
Write example using `CloudBucketMount` with LoRAs

luiscape:
This has an example that shows how to download multiple LoRA adapters from HuggingFace Hub into an S3 bucket (using `CloudBucketMount`) and then load adapters from that bucket during inference.


erikbern:
lol @ the ikea instructions

PR #648, opened 2024-03-12
Update examples to use @web_server

ekzhang:
New feature introduced in https://github.com/modal-labs/modal-client/pull/1513

This removes usage of our `asgiproxy` fork.

Also: @aksh-at mentioned in passing that the "vLLM" example also can be updated? Right now not sure which one this is though.


erikbern:
very cool!

PR #655, opened 2024-03-18
add web endpoint, code cleanup

kning:
* Move checkpoints / plugins config directly as global variables
* Add web endpoint in addition to function
* (hopefully) clearer documentation on what files / functions do what in the process; i don't think this has to be super complete since this code will mainly function as a companion to a blog post that will tie the pieces together more succinctly

Things I tried, but decided against:
* @ekzhang suggested it'd be cleaner for workflow_api.py to directly import the generated python workflow; unfortunately the generated one isn't super usable as is so decided to also leave it out of git and recommend using it simply as reference

Next steps:
* I want to try a more complicated workflow (this is basically just the vanilla demo version) that has custom checkpoints / plugins and see if there are edge cases in this workflow I discover through that process. Also think a more exciting example is in order for the blog post.

### Type of Change

- [ ] New example


06_gpu_and_ml/comfyui/comfy_api.py
```diff
@@ -117,7 +121,15 @@ def query_comfy_via_api(workflow_data: dict, prompt: str, server_address: str):
     return image_list
 
 
-@stub.function(image=image, gpu="any")
+@stub.function(
+    image=image,
+    gpu="any",
+    mounts=[
+        modal.Mount.from_local_file(
+            comfyui_workflow_data_path, "/root/workflow_api.json"
```
ekzhang, line 129: I would suggest passing the file's contents as a string argument rather than a mount here!

PR #666, opened 2024-03-26
parallelize typechecking so it runs faster, fix failing examples

charlesfrye:
resolves a TODO and enables us to bring back typechecking without sacrificing latency of CI/CD on example PRs

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -1,9 +1,10 @@
 name: Typecheck
 on:
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


charlesfrye:
>3x faster! 🎉 

at ~50s it is now almost as fast as the pytest check, which is 30s


Screenshot 2024-03-26 at 1 33 13 PM

charlesfrye:
note that it will also start failing if the main typechecking component starts taking longer than 60s!

PR #669, opened 2024-03-26
remove many old examples, move some into misc with reduced QoS

charlesfrye:
Some of the examples are getting a bit dated. Additionally, we'd like to focus on a smaller set of high quality examples so that we can be sure folks' first experience of Modal is good without creating too high of a maintenance burden.

This PR removes a number of dated or superseded examples. It further moves several of them into the `misc` folder, which now has a README that identifies it as a set of examples with lower QoS. It also removes `misc` from the oversight of the synthetic monitoring system.

### Type of Change

- [ ] New example


ekzhang:
> It also removes `misc` from the oversight of the synthetic monitoring system.

Great idea! Sounds like it would help us focus on monitoring examples that matter that are more widely used.

charlesfrye:
converting to draft to prevent accidental merge before the frontend is ready

PR #671, opened 2024-03-27
Fix typos in comfyui / a1111 examples

ekzhang:
Fix a couple minor typos.


charlesfrye:
good catch! yeah, that timeout was supposed to be 180

PR #693, opened 2024-04-09
removes extra copies of falcon inference examples

charlesfrye:
These were supposed to be removed in #669


06_gpu_and_ml/llm-structured/instructor/inference.py
```diff
@@ -0,0 +1,77 @@
+# # Fast inference with vLLM (Mistral 7B)
```
mwaskom, line 1: Am I missing something here? Looks like this is demonstrating vLLM inference as an openAI-compatible server, not anything to do with `instructor`?

Also we've got a lot of vLLM copy-pasta going on, I'm making updates in one place while we're spawning copies of the old version elsewhere :D. We need a better approach here. (Hard problem though!)

PR #697, opened 2024-04-10
Updates to mixtral vllm example

mwaskom:
Follow-on to #691 in the Mixtral example. Mostly the same updates as there, but also some changes to the local entrypoint logic. Because we call `flush=True`, we were getting a single token on each line when we streamed the output into the terminal:

image

But if you remove `flush=True`, you don't get any streaming.

As a compromise, I am flushing every time a token ends with a newline. That still mostly gives you the "streaming" effect while keeping the text in the terminal readable.

### Type of Change

- [ ] New example


charlesfrye:
i like the compromise!

PR #698, opened 2024-04-12
Replace keep_warm with container_idle_timeout in example

freider:
Keep warm can easily lead to a budget getting consumed by an unknowing user who deploys this to play around with it and then forgets it.


thundergolfer (approved):
Makes sense. The benefit over `container_idle_timeout` is that `keep_warm` immediately spins up the container whereas the former waits for a request. This shortens the user perceived cold start a bit. But tradeoff probably not worth it. We should just speed up cold start 🙂

PR #702, opened 2024-04-18
removes blender example

charlesfrye:
This example is no longer working due to a system package issue. Most of it hasn't been touched in two years and the setup is sufficiently janky that it feels like a bad example of what the platform is capable of.

But if someone wants to resurrect the example post-deletion and make it more stable, they're welcome to!


charlesfrye:
we're keeping the example

PR #713, opened 2024-04-22
Update deploy and invoke section for vllm mixtral example

aastroza:
Fix outdated example reference and update with current working code.

### Type of Change

- [ ] New example


charlesfrye:
thanks!

aastroza:
Hey @charlesfrye ! I'm still exploring this example and I've realized that the `ray==2.10.0` dependency only works with `python<=3.10`. Do you think it's worth adding something like `python_version="3.10"` in the `Image` declaration? Because whether the example works might depend on the local python version, I think.

PR #715, opened 2024-04-24
Add python version to the image to prevent pip install error

jung0072:
### Short summary of the PR
There was an error when I tried the example. Since my local python version was 3.12.2, ray package with version 2.10 was not supported throwing an error:
```
ERROR: Could not find a version that satisfies the requirement ray==2.10.0 (from versions: none)
ERROR: No matching distribution found for ray==2.10.0
```
I added a python version to the image to resolve this error. 

### Type of Change
Example updates (Bug fixes, new features, etc.)


charlesfrye:
Thanks for catching that!

PR #716, opened 2024-04-24
chore: bump ruff to 0.4.1

chenrui333:
### Type of Change

- [ ] New example

## Outside contributors

bump ruff to the latest as it is faster, see https://astral.sh/blog/ruff-v0.4.0

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,7 +19,7 @@ jobs:
           python-version: "3.11"
 
       # keep version here in sync with .pre-commit-config.yaml and other modal repos
-      - run: pip install ruff==0.2.1
+      - run: pip install ruff==0.4.1
 
       - run: ruff check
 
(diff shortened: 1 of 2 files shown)


charlesfrye:
thanks for the contribution!

we try to keep the ruff version consistent between our internal repos and this repo, as noted in [this comment](https://github.com/modal-labs/modal-examples/blob/1a2d861649ff70cb728a56bedf432e1d12cde4e8/.pre-commit-config.yaml#L3).

internally, we're on `0.2.1`, so can't merge this just yet. but let me take a look!

chenrui333:
I see, make sense, but definitely worth adopting ruff 0.4.0+ though :)

charlesfrye:
The speedup is nice, and we love making things faster at Modal!

But `ruff` linting is generally not the bottleneck for us at any step of the QA process -- in-editor, pre-commit, or CI. And the way you make a system faster is by focusing your resources for improvement on the bottlenecks!

PR #718, opened 2024-04-25
Edit hello world example

ekzhang:


01_getting_started/hello_world.py
```diff
@@ -15,20 +15,17 @@
...
 # Modal takes code and runs it in the cloud.
 #
 # So first we've got to write some code.
 #
-# Let's write a simple function:
-# log `"hello"` to standard out if the input is even
-# or `"world"` to standard error if it's not,
-# then return the input times itself.
+# Let's write a simple function that takes an integer and
+# prints a message based on whether its value is even or odd. Then,
```
charlesfrye, line 27: Why remove detail here?

01_getting_started/hello_world.py
```diff
@@ -49,13 +46,13 @@ def f(i):
 #
 # Now let's see three different ways we can call that function:
 #
-# 1. As a regular `local` call on your computer, with `f.local`
+# 1. As a regular `local` call on your computer, with `f.local()`
 #
-# 2. As a `remote` call that runs in the cloud, with `f.remote`
+# 2. As a `remote` call that runs in the cloud, with `f.remote()`
 #
-# 3. By `map`ping many copies of `f` in the cloud over many inputs, with `f.map`
```
charlesfrye, line 56: The use of backticks around `map` here is intentional -- it is symmetric with `local` and `remote` in the other two bullet points.

The point is to help the reader connect the specific Modal objects (`.remote`, `.local`, `.map`) with concepts they already understand (remote and local execution, mapping a function over many inputs).

01_getting_started/hello_world.py
```diff
@@ -49,13 +46,13 @@ def f(i):
...
+# 3. By _mapping_ many copies of `f` in the cloud over a sequence of inputs, with `f.map()`
```
charlesfrye, line 53: [Why waste time say lot word when few do trick](https://media1.tenor.com/m/IsYdPRq7bjcAAAAC/why-waste-time-when-few-word-do-trick.gif)?

I don't think any target reader would be take the non-technical term "many" to mean we need `.map` to call a function with multiple inputs.

01_getting_started/hello_world.py
```diff
@@ -86,16 +83,16 @@ def main():
...
-# **in the cloud**, on Modal's infrastructure, not locally on our computer.
+# _in the cloud_, on Modal's infrastructure, not locally on our computer.
 #
 # In short, we took the function `f`, put it inside a container,
 # sent it the inputs, and streamed back the logs and outputs.
 #
 # ## But why does this matter?
 #
-# Try doing one of these things next to start seeing the full power of Modal!
+# Try doing one of these things next to see the full power of Modal.
```
charlesfrye, line 93: Why imply that the full power is subsumed by doing just one of the things below?

And also, I think that talking about the "full power" of our platform is an apposite time for an exclamation point.

01_getting_started/hello_world.py
```diff
@@ -86,16 +83,16 @@ def main():
...
 #
-# ### You can change the code and run it again
+# ### Change the code and run it again
```
charlesfrye, line 95: I intentionally changed these to be more active and to map onto the question in the section title above: "Why does this matter?".

It matters _because_ you can change the code and run it again -- more quickly and easily than another platform.

It matters _because_ you can map over more data. It matters _because_ you can run something more interesting.

01_getting_started/hello_world.py
```diff
@@ -106,21 +103,20 @@ def main():
...
-# ### You can run a more interesting function
+# ### Run something more interesting
 #
-# The function `f` is obviously silly and doesn't do much, but in its place
+# The function `f` is silly and doesn't do much, but in its place
 # imagine something that matters to you, like:
 #
-# * Running [language model inference](/docs/examples/vllm_mixtral) or [fine-tuning](/docs/examples/slack-finetune)
-# * Manipulating [audio](/docs/examples/discord-musicgen) or [images](stable_diffusion_xl_turbo)
+# * Deploying [language model inference](/docs/examples/vllm_mixtral) or [fine-tuning](/docs/examples/slack-finetune) workloads.
```
charlesfrye, line 118: I wouldn't say that fine-tuning is a "workload" I want to "deploy".

And without a broader discussion of deployment as distinct from function execution, "running inference" doesn't imply to me that it can't be deployed.

I can see an argument for including deployment in the Modal hello world example, but that's not what we're doing here. 

More broadly, I don't see what "workloads" does for us here.

01_getting_started/hello_world.py
```diff
@@ -15,20 +15,17 @@
 
 import modal
 
-app = modal.App(
-    "example-hello-world"
-)  # Note: prior to April 2024, "app" was called "stub"
+app = modal.App("example-hello-world")
```
charlesfrye, line 18: I was initially resistant to removing this comment -- what if new users shortly encounter `Stub` and get confused -- but on further reflection I think we should just remove it now. There are other places where they can learn that information, and the hello world should be clean.

01_getting_started/hello_world.py
```diff
@@ -49,13 +46,13 @@ def f(i):
 #
 # Now let's see three different ways we can call that function:
 #
-# 1. As a regular `local` call on your computer, with `f.local`
+# 1. As a regular `local` call on your computer, with `f.local()`
```
charlesfrye, line 49: Happy to accept this change, but want to note that to me, `f.local()` reads as "the _output_ of invoking the function `f.local()`", not as "the function `f.local` itself".

I've heard both opinions from others, but outside of a functional programing context, the use of parens to denote that the object being discussed is a function is more popular.

01_getting_started/hello_world.py
```diff
@@ -106,21 +103,20 @@ def main():
...
 #
-# * Running [language model inference](/docs/examples/vllm_mixtral) or [fine-tuning](/docs/examples/slack-finetune)
-# * Manipulating [audio](/docs/examples/discord-musicgen) or [images](stable_diffusion_xl_turbo)
+# * Deploying [language model inference](/docs/examples/vllm_mixtral) or [fine-tuning](/docs/examples/slack-finetune) workloads.
+# * Manipulating [audio](/docs/examples/discord-musicgen) or [images](/docs/examples/stable_diffusion_xl_turbo).
 # * [Collecting financial data](/docs/examples/fetch_stock_prices) to backtest a trading algorithm.
 #
-# Modal lets you parallelize that operation effortlessly by running hundreds or
-# thousands of containers in the cloud.
+# Modal lets you parallelize these operations effortlessly by running hundreds of containers in the cloud.
```
charlesfrye, line 122: Why drop the "thousands"?

charlesfrye (changes requested):
Thanks for cleaning up some obvious nits -- the relative link on SDXL, the missing comma.

I disagree with some of the stylistic changes, so I would rather not merge this PR until we come to consensus.

01_getting_started/hello_world.py
ekzhang: It's a bit wordy. Also you can't actually do this. :)

01_getting_started/hello_world.py
ekzhang: image

"Try doing one of these things next!" ->
"You can change the code"

01_getting_started/hello_world.py
ekzhang: Just trying to flow better into the next paragraph: less words, and don't stop here

01_getting_started/hello_world.py
ekzhang: Trying to explain what a map is (a sequence of inputs), not everyone knows what map is

01_getting_started/hello_world.py
ekzhang: Not everyone knows what a map is

01_getting_started/hello_world.py
ekzhang: Will edit it again. Inference and fine-tuning are quite different kinds of workloads and I think it missed the mark

charlesfrye:
incorporated the feedback we were aligned on into #872, so closing this PR

PR #720, opened 2024-04-26
python_version in Image declaration

aastroza:
I've realized that the `ray==2.10.0` dependency only works with `python<=3.10`. Added `python_version="3.10"` to avoid `pip install` errors.

### Type of Change

- [ ] New example


charlesfrye:
thanks again Alonso!

PR #724, opened 2024-04-29
Add Nerf Studio Example

nbardy:
This adds a nerf studio example to modal. 

Setting this up for myself and the build setup got a little complicated. 

### Type of Change

- [X] New example

## Checklist

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/nerf_studio.py
+++ b/06_gpu_and_ml/nerf_studio.py
@@ -41,8 +41,10 @@ image = (
     )
     .run_commands(
         "pip install ninja git+https://github.com/NVlabs/tiny-cuda-nn/#subdirectory=bindings/torch",
-        "git clone https://github.com/nerfstudio-project/nerfstudio.git",
-        "cd nerfstudio && pip install -e .",
+    )
+    .run_commands(
+        "git clone https://github.com/nerfstudio-project/nerfstudio.git /nerfstudio",
+        "cd /nerfstudio && pip install -e .",
         gpu=gpu_device,
     )
     .pip_install(
@@ -67,5 +69,5 @@ app = modal.App(name="nerfstudio-on-modal", image=image)
 def nerfstudio_web():
     import subprocess
 
-    cmd = "cd nerfstudio && python -m nerfstudio.server --port 7007"
+    cmd = "cd /nerfstudio && python -m nerfstudio.server --port 7007"
     subprocess.Popen(cmd, shell=True)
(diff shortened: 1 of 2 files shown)


06_gpu_and_ml/nerf_studio.py
```diff
@@ -0,0 +1,73 @@
...
+image = (
+    modal.Image.conda()
+    .apt_install("git")
+    .run_commands(
+        "apt-get update",
+        "apt-get install -y wget build-essential git",
+        "pip install --upgrade pip",
+    )
+    .conda_install(
+        "nvidia::cuda-toolkit=12.4.1",
```
charlesfrye, line 16: I'm kinda surprised this works? The host driver is 12.2.

Is a later step maybe installing the CUDA toolkit at <=12.2?

charlesfrye:
Would love to get a NeRF example in!

The install does seem kinda messy though.

Looking at [their docs](https://github.com/nerfstudio-project/nerfstudio/blob/main/docs/quickstart/installation.md#dependencies), I see

> pip install torch==2.1.2+cu118 torchvision==0.16.2+cu118 --extra-index-url https://download.pytorch.org/whl/cu118

Torch now bundles the CUDA toolkit, including nvcc, so the above should work to get you everything. Maybe it's just a matter of setting CUDA_HOME and symlinking?

I also don't think we to install the [editable version](https://github.com/modal-labs/modal-examples/pull/724/files#diff-bfb14c16c6f7a0f93aa27d394e670a613b76a3ab6c9c6d4d6bc195e61deff0c9R45-R49) of the library, since folks won't be editing the library on Modal.

So I think we can shrink this down to a few `pip_install`s -- one for main deps (torch, transformers, ninja), one for tiny-cuda-nn, one for nerfstudio, and maybe one more for [kplanes-nerfstudio](https://docs.nerf.studio/nerfology/methods/kplanes.html).

HMU in the Modal Slack if you want to pair on it for a bit.

nbardy:
`dromni/nerfstudio`

I ended up finding a docker image. I think this is probably the better way to do it.

> I also don't think we to install the [editable version](https://github.com/modal-labs/modal-examples/pull/724/files#diff-bfb14c16c6f7a0f93aa27d394e670a613b76a3ab6c9c6d4d6bc195e61deff0c9R45-R49) of the library, since folks won't be editing the library on Modal.

I do want to fork the repo and develop on it from this setup. Probably should move that out to an optional flag to use a local version.

@charlesfrye 
Hoping to pick this up in the evenings after work this week. But not so much time this week.

charlesfrye:
@nbardy closing this one for now but keeping the branch around in case you want to pick it up again!

PR #735, opened 2024-05-05
Run examples on change

charlesfrye:
This PR adds a new workflow, Run, that detects changed examples and runs them. This gets us another step closer to full continuous integration of example changes -- it will reduce the number of issues that show up via monitoring (🎉), but we're still missing isolated builds.

To keep testing fast and isolated, all changed examples are executed in parallel job runs (via a jobs `matrix` output by the [`changed-files` action](https://github.com/tj-actions/changed-files)).

I also centralized the environment setup for the different CI workflows into a single action file, as in the main repo.

FYI, I would not have been able to get this working without [the `act` tool](https://github.com/nektos/act), which runs GitHub Actions locally. Really speeds up debugging!

### Type of Change


charlesfrye:
Had to revert the updates to typechecking.

`mypy<0.981` cannot typecheck any code with positional-only parameters, including `modal`, due to a [bug](https://github.com/python/mypy/issues/13627). In the old way of setting up the type-checking environment, `modal` was not installed, and so it was not checked.

I tried to update `mypy`, but versions with the fix are much, much [slower](https://github.com/python/mypy/issues/14867). As configured, typechecking was taking about two minutes _per example_, instead of under a minute total.

So I went back to the old way.

Note that we _also_ can't effectively typecheck any code running inside Modal functions, because the dependencies aren't available to `mypy`.

So we are back to neither type-checking the core code of example apps nor the code from our library.

charlesfrye:
merging but still welcome C+C/review to improve!

thundergolfer (commented):
LGTM. thanks for wading through all that YAML 😄

PR #740, opened 2024-05-10
Specify bpy / Blender 4.1.0

ekzhang:
> Each Blender release supports one Python version, and the package is only compatible with that version.

Just realized we should probably pin this dependency to avoid future breakage. Blender 4.1.0 happens to match Python 3.11.


ekzhang:
Kind of wacky to see 10 GPUs chugging along doing ray tracing in CI

charlesfrye:
@ekzhang we could add `FRAME_SKIP` as a CLI arg and adjust the frontmatter command?

but this CI only runs on changes to the example, so it's probably only adding cost occasionally and on top of the cost to edit and debug, rather than adding cost frequently and from 0

ekzhang:
Yeah to be clear I wasn't saying the cost is a problem! :)

PR #742, opened 2024-05-14
expand comfyui example

kning:
This expands the existing ComfyUI example to include things like running via API and optimizing inference performance.

This kind of breaks `comfy_api.py`, but figure we should move to this as the definitive guide for ComfyUI plus Modal anyway. The one thing `comfy_api.py` has that could be worth keeping is the `get_python_workflow` piece. Wasn't able to really get it working 

### Type of Change

- [ ] New example

## Checklist

## Outside contributors


06_gpu_and_ml/comfyui/comfy_ui.py
```diff
@@ -1,140 +1,157 @@
...
+
+    def _run_comfyui_server(self, port=8188):
+        cmd = f"python main.py --dont-print-server --listen --port {port}"
+        subprocess.Popen(cmd, shell=True)
+
+    @modal.web_server(8188, startup_timeout=30)
+    def ui(self):
+        self._run_comfyui_server()
+
+    # When you run `modal serve 06_gpu_and_ml/comfyui/comfy_ui.py`, you'll see a `ComfyUI.ui` link to interactively develop your ComfyUI workflow that has the custom checkpoints/plugins loaded in.
```
charlesfrye, line 112: note that the linter added spaces in front of the comments, so they aren't rendered properly

kning: what should we do about the ruff ci check then?

charlesfrye: unfortunately, we just have to reorganize the text.

i have looked into ways around this, e.g. adding methods to classes after they are defined, but haven't been able to find a way to do it

PR #762, opened 2024-06-05
stub -> app

thliang01:
### Type of Change

## Checklist

## Contributors

change notebook stub --> app


charlesfrye (commented):
Thanks for catching this!

PR #787, opened 2024-06-19
Fix sglang example after unpinned transitive dependencies broke

freider:
Fixes:
* Upgrade sglang to 0.1.17 which pins the vllm dependency (it was unpinned in 0.1.16 so it broke when vllm was updated)
* Downgrade numpy (unpinned dependency of the outlines package) to pre-2.0 release which is required by outlines in the current version.
* Change the typoed log_evel to log_level to work with sglang 0.1.17

Nits:
* Rename env var GPY_CONFIG to GPU_TYPE to be in sync with var names
* Remove explicit dependencies on unpinned "ninja", "packaging" packages
* Remove the `async` designators on the non-async methods (probably didn't cause issues as is, but could potentially do that since the methods as is are locking the user-code event loop when running, so concurrent inputs would break etc.)

### DIFF
--- a/06_gpu_and_ml/llm-serving/sgl_vlm.py
+++ b/06_gpu_and_ml/llm-serving/sgl_vlm.py
@@ -31,11 +31,11 @@ import requests
 # LLMs are already hard to run effectively on CPUs, so we'll use a GPU here.
 # We find that inference for a single input takes about 3-4 seconds on an A10G.
 #
-# You can customize the GPU type and count using the `GPU_CONFIG` and `GPU_COUNT` environment variables.
+# You can customize the GPU type and count using the `GPU_TYPE` and `GPU_COUNT` environment variables.
 # If you want to see the model really rip, try an `"a100-80gb"` or an `"h100"`
 # on a large batch.
 
-GPU_TYPE = os.environ.get("GPU_CONFIG", "a10g")
+GPU_TYPE = os.environ.get("GPU_TYPE", "a10g")
 GPU_COUNT = os.environ.get("GPU_COUNT", 1)
 
 GPU_CONFIG = f"{GPU_TYPE}:{GPU_COUNT}"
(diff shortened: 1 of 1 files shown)


charlesfrye:
> Remove explicit dependencies on unpinned "ninja", "packaging" packages
> Removed the flash-attn dependency as I couldn't see it actually used (?). Removal didn't seem to affect performance

The Flash Attention algorithm is a performance optimization for attention calculations, so the flash-attn library isn't used directly, it's just detected by libraries and used when present. The package is a PITA to install but used to be the only implementation of FA. `ninja`, `packaging`, and `wheel` need to be installed _before_ `flash-attn` is, and then it needs to be installed with `no-build-isolation`.

Looks like vllm has [an easier-to-install version](https://github.com/modal-labs/modal-examples/actions/runs/9580448381/job/26415109011#step:4:125) so it's no longer needed. That's also why we were even using `from_registry` in the first place, so we can simplify the install here even further!

freider:
> > Remove explicit dependencies on unpinned "ninja", "packaging" packages
> > Removed the flash-attn dependency as I couldn't see it actually used (?). Removal didn't seem to affect performance
> 
> The Flash Attention algorithm is a performance optimization for attention calculations, so the flash-attn library isn't used directly, it's just detected by libraries and used when present. The package is a PITA to install but used to be the only implementation of FA. `ninja`, `packaging`, and `wheel` need to be installed _before_ `flash-attn` is, and then it needs to be installed with `no-build-isolation`.
> 
> Looks like vllm has [an easier-to-install version](https://github.com/modal-labs/modal-examples/actions/runs/9580448381/job/26415109011#step:4:125) so it's no longer needed. That's also why we were even using `from_registry` in the first place, so we can simplify the install here even further!

Ah, neat! I suspected it was something automatically picked up by other packages, but since I didn't notice any performance regression I assumed it wasn't used here (I guess it was used via the bundled dependency you pointed out). Nice cleanup in the subsequent PR!

PR #794, opened 2024-06-21
reword "our computer"

charlesfrye:
The document is written in a collaborative second-person plural, but "our computer" still scans to some readers as referring to Modal's infrastructure.


charlesfrye:
@prbot automerge

PR #799, opened 2024-06-27
add warning about ui

kning:
This happened to a user so just adding note here to remember to close ComfyUI tabs to avoid charges. Also linking to popular volume gist.

### Type of Change

- [ ] New example


kning:
@prbot approve

PR #816, opened 2024-07-19
example: modal llm inference from dbt python model

kramstrom:
### Type of Change

- [x] New example

## Checklist

## Outside contributors


06_gpu_and_ml/llm-serving/trtllm_llama.py
```diff
@@ -567,7 +567,7 @@ def main():
 
 class GenerateRequest(pydantic.BaseModel):
     prompts: list[str]
-    settings: Optional[dict]
+    settings: Optional[dict] = None
```
kramstrom, line 570: Added `= None` as otherwise this wasn't really optional

charlesfrye: In Python typing, `Optional` means "can be `None`" and is distinct from "optional argument", i.e. argument with a default.

charlesfrye:
looks like there's a few unrelated deprecation errors from client changes -- e.g. https://github.com/modal-labs/modal-examples/actions/runs/10014558567/job/27684478855?pr=816#step:4:314

06_gpu_and_ml/llm-serving/trtllm_llama.py
kramstrom: True, I guess we could change the `curl` example doc below instead to include a `settings = None` param if you prefer. With fastapi v2 either this has to be `None` or we need to pass it explicitly in the request, so one of the two places needs an update

kning:
very cool! i also have this dlt modal example running on our data currently: https://github.com/modal-labs/modal-etl/blob/main/dlt/postgres_pipeline.py. i think could be a great blog post to combine those two into one e2e etl example.

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
+# ---
+# cmd: ["modal", "run", "10_integrations/dbt_modal_inference/duck_db_modal_inference.py"]
```
charlesfrye, line 2: this shouldn't be necessary, right? -- this should be the default command for anything testable

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
...
+# ---
+# # dbt python model with llm inference
+#
+# In this example we demonstrate how you could combine dbt python models with llm
```
charlesfrye, line 6: I think it's a bit confusing to use "models" multiple ways here. Are there alternative ways to describe what DBT is doing? Is it really just providing data models?

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
...
+#
+# ## Overview
+#
+# This example runs [dbt](https://docs.getdbt.com/docs/introduction) with a [duckdb](https://duckdb.org)
+# backend directly on top of modal, but could be translated to run on any dbt-compatible
+# database that supports python models. Similarly you could make these requests from UDFs
+# directly in SQL instead if you don't want to use python models.
+#
+# In this example we use an llm deployed in a previous example: [Serverless TensorRT-LLM (LLaMA 3 8B)](https://modal.com/docs/examples/trtllm_llama)
+# but you could easily swap this for whichever modal function you wish. We use this to classify the sentiment
```
charlesfrye, line 17: nit: we generally capitalize Modal and Function -- to separate the proper noun/product component "Function" from the normal old functions in Python

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
...
+# for free-text product reviews and aggregate them in subsequent dbt sql models.
```
charlesfrye, line 18: Did you also generate the reviews with the LLM? That's a neat point to include and justifies the use of a larger model.

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
...
+#
+#
+# ## dbt & image configuration
+#
+# We set up the environment variables necessary for dbt and
+# create a slim debian and install the packages necessary to run.
```
charlesfrye, line 24: See the [updated description](https://github.com/modal-labs/modal-examples/commit/da92247c52fd5a6e32f9cbad8567704ca733a950) of a similar section of the other DBT example for a sample of how we prefer to explain the environment setup.

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
```diff
@@ -0,0 +1,276 @@
...
+# └──────────────────────────────────────────┴──────┴───────────────────────┴─────────┘
+# ```
+#
+# ## Python dbt model
+# The python dbt model in `dbt_modal_inference_proj/models/product_reviews_sentiment.py` is quite simple.
+# It configures dbt to store the results locally in parquet, fetches the inference url,
+# uses a batch reader to iterate over the inputs and call the `batcher` method
+# and adds a new field from the LLM
+#
+# ```python
```
charlesfrye, line 200: Altogether, there's quite a bit of inlined Python here and we would need to duplicate any changes. Do we need the code here or can we point readers to the files, where we can put comments?

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
kramstrom: I agree it's a bit confusing, but calling them sql / python / data models really is the dbt way. I can see if I can make it a bit clearer though

10_integrations/dbt_modal_inference/duck_db_modal_inference.py
kramstrom: That would work, do you think we should link to the files on github directly?

PR #824, opened 2024-07-26
adds yolo fine-tuning example

erik-dunteman:
### Type of Change

- [x] New example

## Checklist

## Outside contributors


06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+    from ultralytics import YOLO
+    # Choose the large size checkpoint
+    # Other model sizes can be found at: https://docs.ultralytics.com/models/yolov10/#performance
+    model = YOLO('yolov10l.pt') # This downloads the model's base weights if not already present in the image.
+    return model
+
+yolo_image = (
+    modal.Image.from_registry("pytorch/pytorch:2.3.1-cuda11.8-cudnn8-runtime")
+    .apt_install(["libgl1-mesa-glx", "libglib2.0-0"])
+    .pip_install(["ultralytics", "datasets", "opencv-python"])
```
luiscape, line 42: We like pinning to avoid drift-induced breakage.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+    .run_function(load_yolo_weights)
+
+app = modal.App("train-yolo")
```
luiscape, line 45: It seems that `yolo_image` is the only image here. If so, then we can add it to the app:

```python
app = modal.App(..., image=yolo_image)
```

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+ 
+    for split in ["train", "val", "test"]:
+        print(f"Converting {split} dataset...")
+        
+        ann_dir = os.path.join(dataset_path, f"base_dataset/{split}/annotations")
+        img_dir = os.path.join(dataset_path, f"base_dataset/{split}/images")
+        
+        os.makedirs(root_folder / "labels" / split, exist_ok=True)
+        os.makedirs(root_folder / "images" / split, exist_ok=True)
```
luiscape, line 99: Slight preference for using `Path`-baed operations everywhere instead of `os` etc.

luiscape (commented):
Just a few nits. Thanks for the PR! ❤️

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
+# ---
+# runtimes: ["runc", "gvisor"]
```
charlesfrye, line 2: fyi, we no longer use the `runtimes` flag. all workloads run on `gvisor` by default

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+# ---
+#
+# # Training and Inferencing YOLOv10
```
charlesfrye, line 5: Slight preference to avoid  the"Inferencing" gerund in favor of "Inference" or "Running"

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+# Example by [@AnirudhRahul](https://github.com/AnirudhRahul)
+#
+# [YOLOv10](https://docs.ultralytics.com/models/yolov10/) is the most modern version of the popular "You Only Look Once" ("YOLO") computer vision model, 
+# which has been trained to perform bounding box object detection.
+#
+# In this example, we use the V10 model via the [Ultralytics package](https://docs.ultralytics.com/). We will:
+# - Fine tune on a custom dataset, in our case the DoclayNet dataset for object detection within documents.
+# - Deploy an inference endpoint.
+
+# For commercial use, sure to consult the [ultralytics software license](https://docs.ultralytics.com/#yolo-licenses-how-is-ultralytics-yolo-licensed).
```
charlesfrye, line 16: At first I was bummed, assuming they only had a bespoke license, but the AGPL is OSI-compliant! Worth calling out in this era of sus licensing.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
+# ---
+# runtimes: ["runc", "gvisor"]
+# ---
+#
+# # Training and Inferencing YOLOv10
+#
+# Example by [@AnirudhRahul](https://github.com/AnirudhRahul)
+#
+# [YOLOv10](https://docs.ultralytics.com/models/yolov10/) is the most modern version of the popular "You Only Look Once" ("YOLO") computer vision model,
```
charlesfrye, line 9: I would avoid "most modern", since it rots quickly and unexpectedly. Can we frame this differently? E.g. "The popular YOLO model lineage provides high-quality object detection in an economical package. In this example, we use the YOLOv10 model, released on Month Day, Year".

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+# - Fine tune on a custom dataset, in our case the DoclayNet dataset for object detection within documents.
+# - Deploy an inference endpoint.
+
+# For commercial use, sure to consult the [ultralytics software license](https://docs.ultralytics.com/#yolo-licenses-how-is-ultralytics-yolo-licensed).
+
+# ## Building Our Image
+#
+# The example uses the Ultralytics package to run training inference and the Huggingface dataset package to load DoclayNet data.
+# 
+# All packages are installed into a Pytorch 2.3.1 base image
```
charlesfrye, line 22: Any reason we can't use `debian_slim`? I imagine it's because ultralytics wants a `CUDA_HOME`, whereas PyTorch installation via `pip` does a `site-packages` install.

If so, we should at least prefer the NVIDIA Docker container that we recommend in the guide and install PyTorch on top.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+# In addition, we download the base YOLO model weights into the container at build-time
+# using the `run_function` function
+
+import modal
+import urllib.request, os, yaml, json, shutil, zipfile
+from pathlib import Path
+
+def load_yolo_weights():
+    from ultralytics import YOLO
+    # Choose the large size checkpoint
```
charlesfrye, line 34: Any comments on why this size?

For example: even if we turn this into a webcam demo using websockets, I don't think we actually care about the extra few ms of latency from XL/L vs S/N. Good to call out latency numbers, so folks know how to adjust this example.

06_gpu_and_ml/yolo/yolo.py
charlesfrye: @erik-dunteman see the PR template for a detailed policy.

`apt_install`ed libs are generally not pinned, because they are more dependable than Python libraries.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+
+app = modal.App("train-yolo")
+
+# ## Downloading The Dataset
+# 
+# Similar to downloading model weights at build-time, it may be tempting to also download the DoclayNet dataset at build-time.
+# We instead choose to download the dataset into a Modal Volume, using a Modal Function.
+# This approach has two benefits:
+# - Reduces the size of the container image, making cold boots much faster. This is especially important for inference when there's no need for having the data in the container.
+# - Makes it easier to train on many datasets, storing to create many finetunes usable for inference.
```
charlesfrye, line 53: This is a great callout. Would love to have multiple finetunes on the Volume.

Imagining a webcam demo where there are finetunes for face detection and for product detection and so on, and the user can switch between them.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+    from tqdm import tqdm
+    class DownloadProgressBar(tqdm):
+        def update_to(self, b=1, bsize=1, tsize=None):
+            if tsize is not None:
+                self.total = tsize
+            self.update(b * bsize - self.n)
+    with DownloadProgressBar(unit='B', unit_scale=True, miniters=1, desc=url.split('/')[-1]) as t:
+        urllib.request.urlretrieve(url, filename=output_path, reporthook=t.update_to)
+
+def convert_dataset_to_yolo(dataset_path, output_path):
```
charlesfrye, line 68: Surprised at the length of this code -- is there not a utility we can pull from ultralytics?

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+
+                    annotation = (category, center_x, center_y, width, height)
+                    if annotation not in seen_annotations:
+                        f.write(f"{category} {center_x} {center_y} {width} {height}\n")
+                        seen_annotations.add(annotation)
+
+@app.function(
+    image=yolo_image,
+    volumes={"/root/data/": volume},
+    cpu=4,
```
charlesfrye, line 137: We generally don't request lots of CPU on our examples, because jobs can usually burst up to consume more CPU when they need it. Did you find that this example required CPU pinning for some reason?

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+            
+            # Save the results
+            output_path = os.path.join(output_dir, os.path.basename(image_file))
+            print("SAVING: ", output_path)
+            results[0].save(output_path)
+
+        print(f"Inference complete. Results saved to {output_dir}")
+    
+@app.local_entrypoint()
+def infer(model_name = "doclaynet-base", input_dir = "/root/data/doclay-dataset/images/val"):
```
charlesfrye, line 258: I generally prefer the `local_entrypoint` to run on its own, without requiring any other functions like `download_dataset` or `train` to run.

charlesfrye: And we call the `local_entrypoint` by `main` usually, because that's the purpose it serves.

06_gpu_and_ml/yolo/yolo.py
```diff
@@ -0,0 +1,288 @@
...
+# With the dataset downloaded, we proceed to building our training pipeline.
+
+@app.function(
+    image=yolo_image,
+    timeout=1500,
+    volumes={"/root/data/": volume},
+    gpu='a100',
+    cpu=2,
+)
+def train(model_name = "doclaynet-base", data_path = '/root/data/doclay-dataset/data.yaml'):
```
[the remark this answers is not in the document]
charlesfrye: logging in terminal is okay for now. ultralytics dropped wandb support

PR #846, opened 2024-08-18
fix parsing for stock tickers, update text

charlesfrye:


thundergolfer:
Thanks! Was just about to do this myself.

PR #877, opened 2024-09-16
update code_interpreter, minor text fixes, make testable

charlesfrye:
main change: added an `App` to the `Sandbox` since we're deprecating app-less Sandboxes.


thundergolfer (approved):
LGTM, cheers

PR #893, opened 2024-09-27
turn off preview docs on main

charlesfrye:
It fails because there's no issue to comment on.

Could add conditional logic for posting but running the check is not necessary since the action is run on a merge commit already.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8c11d84.modal.run

PR #894, opened 2024-09-27
fixes url in dreambooth example

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4fb57bd.modal.run

PR #898, opened 2024-09-30
fix header

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6f3245b.modal.run

PR #901, opened 2024-10-01
Update comfy example to emphasize how to iterate & avoid redownloading weights

kramstrom:
~Basically adding parts of [this gist](https://gist.github.com/kramstrom/a3784d8246948db647e7e66df980d090) as further improvements to our comfy example. It does mainly two things:~
~- store model weights on a volume so that they can be re-used across different apps and allows you to change your image however you want it without having to re-download everything~
~- it changes the model downloading to use huggingface helpers instead which allows you to pass in a huggingface token and download protected models~

~Not sure if these fit best as inline code here or if the gist should be its own example / guide / blog post or scrapped completely and just remain as a gist.~

Updated the example instead to split up the image building steps into multiple & explain each step more thoroughly and how users should add new models and custom nodes without having to rebuild the whole image (which is why users want to store them on volumes in the first place)

### Type of Change

- [ ] New example

## Checklist

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/comfyui/comfyapp.py
+++ b/06_gpu_and_ml/comfyui/comfyapp.py
@@ -43,8 +43,6 @@
 #
 # ## Setup
 #
-# First, we define the environment we need to run ComfyUI using [`comfy-cli`](https://github.com/Comfy-Org/comfy-cli). This handy tool manages the installation of ComfyUI, its dependencies, models, and custom nodes.
-
 
 import json
 import subprocess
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-cc6c646.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5b2f894.modal.run

charlesfrye (changes requested):
Discussed offline -- we don't want to add this as a block of untested code at the end. Kasper is rewriting to merge this with the rest of the material.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-12cfa19.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2e5b652.modal.run

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -54,33 +52,97 @@
...
 
+# ### Building up the environment
+#
+# ComfyUI setups can be complex, with a lot of custom nodes and models to manage.
+# We'll use [`comfy-cli`](https://github.com/Comfy-Org/comfy-cli) to manage the installation of ComfyUI, its dependencies, models, and custom nodes.
+#
+# We start from a base image and specify all of our dependencies.
+# We'll call out the interesting ones as they come up below.
+#
+# Note that these dependencies are not installed locally
```
charlesfrye, line 63:
```suggestion
# Note that these dependencies are not installed locally.
```

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -54,33 +52,97 @@
...
+# they are only installed once in the remote environment where our app runs,
```
charlesfrye, line 64:
```suggestion
# They are only installed in the remote environment where our app runs.
```

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -54,33 +52,97 @@
...
+# on subsequent runs, the image will be cached and reused.
```
charlesfrye, line 65:
```suggestion
# This happens the first time. On subsequent runs, the cached image will be reused.
```

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -54,33 +52,97 @@
...
+        "comfy --skip-prompt model download --url https://huggingface.co/black-forest-labs/FLUX.1-schnell/resolve/main/flux1-schnell.safetensors --relative-path models/unet"
     )
-    .run_commands(  # download a custom node
+    # Add .run_commands(...) calls for any other models you want to download
+)
+
+# #### Downloading custom nodes
+#
+# We'll download custom nodes using `comfy-cli` too.
+# (alternatively, you can install them by cloning the git repositories to your `/root/comfy/ComfyUI/custom_nodes`
```
charlesfrye, line 109:
```suggestion
# Alternatively, you can install them by cloning the git repositories to your `/root/comfy/ComfyUI/custom_nodes`
```

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -54,33 +52,97 @@
...
+# directory and installing the required dependencies manually.)
```
charlesfrye, line 110:
```suggestion
# directory and installing the required dependencies manually.
```

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8c3c5c7.modal.run

charlesfrye (approved):

kramstrom:
> it looks like here then that we're not publishing the volume-based approach and just telling people to append dependencies to the image definition to leverage caching?

Yeah we opted to go for the current approach with appending dependencies, if people really want to use volumes though we still have the gist to show how it could be done

PR #912, opened 2024-10-03
provide a default value for compile so @build runs

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-e97a585.modal.run

PR #913, opened 2024-10-03
adjust timeout for compilation of flux

charlesfrye:
can take 20 minutes or more -- and timeouts are shared between `enter` phase and main Function


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-840b27b.modal.run

PR #914, opened 2024-10-03
move to decorator pattern, fix incorrect path in text

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c2518e1.modal.run

PR #915, opened 2024-10-04
speed up torch compile with more cacheing

charlesfrye:
we were missing the inductor baseline cache and FX_GRAPH_CACHE ([docs](https://pytorch.org/tutorials/recipes/torch_compile_caching_tutorial.html))

im seeing lower, but highly variable, compile times between ~90s and ~500s, shortened from 1100s


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-7d51d70.modal.run

PR #917, opened 2024-10-04
adds more detail to install_cuda example

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9fa9fbb.modal.run

PR #920, opened 2024-10-04
adds tags for featured examples and use cases

charlesfrye:
to be used by the frontend to determine where examples should be linked from/displayed


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-0f1100d.modal.run

PR #928, opened 2024-10-17
faster checkboxes, wider container

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f17c525.modal.run

PR #931, opened 2024-10-17
better load-testing for checkboxes

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5dc500c.modal.run

PR #945, opened 2024-10-23
switch to a higher param count, lower quant model

charlesfrye:
The new w4a16 quants from Neural Magic are supported in vLLM and give us roughly the same performance and storage size for an 8B model as we had for a w8a8 quant of a 3B model.

### Type of Change

- [ ] New example


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5015ad8.modal.run

PR #948, opened 2024-10-24
faster mochi

charlesfrye:
This PR adds early downloading of the T5 text encoder used by Mochi so that it doesn't need to be downloaded on each cold start.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4605d41.modal.run

erik-dunteman:
👑

PR #951, opened 2024-10-29
Add explicit fastapi installation to all examples that use it

mwaskom:
With the 2024.10 Image Builder, `fastapi` is no longer included in the default container environment and must be explicitly installed. I went through the examples and tried to do that wherever relevant. It seems the best practice is to do `fastapi[standard]` so I mostly tried to do that. In terms of pinning (both of `fastapi` and key transitive dependencies) I tried to match the style of the example I was touching; some "production ready" examples pin extensively, while others are more terse. Note that the current version of fastapi has `pydantic<3` in its requirements, so I don't think we're super likely to get bit by another rough transition there.

Honestly there was a lot to change here and I still might have missed a few, so no promises that these don't fail in CI when we bump the Image Builder version. Note that we will raise informatively during App creation, which is at least better than a surprise runtime error.

### Type of Change

- [ ] New example


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a9da247.modal.run

06_gpu_and_ml/hyperparameter-sweep/hp_sweep_gpt.py
```diff
@@ -106,16 +105,20 @@
...
     "tensorboard==2.17.1"
 )
 
-# And we'll deploy a web UI for interacting with our trained models using Gradio.
 
-ui_image = Image.debian_slim(python_version="3.11").pip_install(
-    "gradio~=4.44.0", "pydantic>=2", "fastapi==0.114.2"
+# We'll serve a simple web endpoint
+web_image = modal.Image.debian_slim(python_version="3.11").pip_install(
+    "fastapi[standard]==0.114.2", "pydantic==2.9.1", "starlette==0.41.2"
```
charlesfrye, line 115: I think we need to either bump the `fastapi` version or downgrade the `starlette` version here, see [this CI failure](https://github.com/modal-labs/modal-examples/actions/runs/11582304105/job/32246787164?pr=951).

charlesfrye (changes requested):
Thanks for the PR!

Not sure why the diffusion examples are failing, but those are less prominent examples, so feel free to merge with failing CI and I will resolve async.

Would prefer to get `dreambooth_app` and `comfy_app` (failed from timeout) and `hp_sweep_gpt` (failed from version incompat) to pass before we merge.

I re-triggered the two timeout failures to see if we can't finish under the time limit now that some of the images/image stages are cached. If not, then run the example's frontmatter `cmd` command in the `modal-labs` workspace outside of CI and open an issue indicating that it times out in CI on build and I'll rework it to run faster. These "failures" aren't user-facing, since users don't have that timeout, so they are not blocking.

mwaskom:
> Not sure why the diffusion examples are failing, but those are less prominent examples, so feel free to merge with failing CI and I will resolve async.

I suspect this is the same issue we were debugging last night: https://modal-com.slack.com/archives/C02ALG62GE7/p1730162610500109

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4c78991.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-69d1421.modal.run

charlesfrye:
The image change revealed an underlying issue with the `comfyapp` example.

This is a user-facing issue, but not caused by this PR. To avoid blocking, I'm going to bypass the branch protections. I've opened #952 to track it.

VDuda:
@charlesfrye 
> `dreambooth_app` and `comfy_app` (failed from timeout)

```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'starlette.requests.Request'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.

For further information visit https://errors.pydantic.dev/2.9/u/schema-for-unknown-type
```

Was wicked interested in trying out this example tonight - here's the error in case it's useful. `06_gpu_and_ml/dreambooth` is the example I was hoping to get up and running. 

Overall incredibly excited to try out modal. Thank you!

PR #955, opened 2024-10-31
removes old vision model training example

charlesfrye:
This example no longer builds -- it spends a long time searching for a compatible `scipy` version, then fails to install it because of missing system libraries (BLAS/LAPACK).

Rather than debugging this, I'm deleting this example. From a quick search, it's no longer linked in the docs. It has been superceded by more recent training examples (`dreambooth_app`, `hp_sweep_gpt`).


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a7c2dfe.modal.run

erikbern:
yay let's delete more!

PR #961, opened 2024-11-05
Return code interpreter example to use exec

pawalt:
Previously we had a bug in exec that made it drop some messages. This bug is fixed so we can return to our preferred exec pattern.

### Type of Change

- [ ] New example

## Checklist

### DIFF
--- a/13_sandboxes/simple_code_interpreter.py
+++ b/13_sandboxes/simple_code_interpreter.py
@@ -67,11 +67,11 @@ def driver_program():
 # that program and execute code in it.
 
 
-def run_code(sb: modal.Sandbox, code: str):
-    sb.stdin.write(json.dumps({"code": code}))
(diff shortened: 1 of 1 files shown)


pawalt:
@prbot automerge

pawalt:
whoops i guess automerge doesn't work on this repo

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b91f452.modal.run

13_sandboxes/simple_code_interpreter.py
```diff
@@ -67,11 +67,11 @@ def driver_program():
 # that program and execute code in it.
```
charlesfrye, line 67: @pawalt could you also update the text here

PR #965, opened 2024-11-07
enable hf_transfer for flux download

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-fa009f0.modal.run

PR #967, opened 2024-11-12
moves class definition into loop, it's slower

charlesfrye:
Move the `class Record` definition into the matmul loop -- it takes longer.

That was the original intent, cf. the text on optimization in "Improving the performance". I moved it out while measuring the difference and forgot to put it back.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4bebe38.modal.run

PR #968, opened 2024-11-12
update tensorboard function config for torch_profiling example

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4430832.modal.run

PR #973, opened 2024-11-14
Bump the version of the tensorflow image in our tensorflow example

mwaskom:
This example runs on Python 3.8, which we're about to drop support for.

I picked an arbitrary "newer but not too new" version which we seem to be able to run the example to completion on. My depth of knowledge is shallow here though! The code doesn't crash and the loss goes down; what more can you ask for? 

### Type of Change


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-727a750.modal.run

PR #975, opened 2024-11-15
stop mounting volume on .cache, just mount huggingface cache

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-7f59531.modal.run

PR #979, opened 2024-11-18
removes instructor embeddings example

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b88be99.modal.run

PR #985, opened 2024-11-19
Use the Node version specified in nvmrc to build

ekzhang:
We're updating from Node 18 to Node 22 LTS. Using the `.nvmrc` file to keep this step consistent and avoid having to change it in multiple places in the future.

This won't work until https://github.com/modal-labs/modal/pull/17482 is deployed.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-005252d.modal.run

charlesfrye:
thanks for the update!

PR #986, opened 2024-11-19
Updates Mochi to use Diffusers

erik-dunteman:
Significantly cheaper inference, faster cold boots, and makes for a much more readable example.

### Type of Change

- [ ] New example

## Checklist

- [ ] Example pins its dependencies- `currently requires diffusers to be installed from source`

### DIFF
--- a/06_gpu_and_ml/text-to-video/mochi.py
+++ b/06_gpu_and_ml/text-to-video/mochi.py
@@ -16,10 +16,9 @@
 # </center>
 
 # Note that the Mochi model, at time of writing,
-# requires several minutes on four H100s to produce
+# requires several minutes on one H100 to produce
 # a high-quality clip of even a few seconds.
-# It also takes a five to ten minutes to boot up.
-# So a single video generation therefore costs about $2
+# So a single video generation therefore costs about $0.33
 # at our ~$5/hr rate for H100s.
 
 # Keep your eyes peeled for improved efficiency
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9b07752.modal.run

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
-tag = f"{cuda_version}-{flavor}-{os_version}"
+app = modal.App()
 
-image = (
-    modal.Image.from_registry(f"nvidia/cuda:{tag}", add_python="3.11")
-    .entrypoint([])
-    .apt_install("git", "ffmpeg")
-    .pip_install("torch==2.4.0", "packaging", "ninja", "wheel", "setuptools")
-    .pip_install("flash-attn==2.6.3", extra_options="--no-build-isolation")
+image = (modal.Image.debian_slim()
```
charlesfrye, line 41: could you set `python_version` to `"3.11"`

charlesfrye, line 41: don't forget to run `ruff`! it's in the requirements under `internal`

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+    .apt_install("git")
+    .pip_install("torch", "accelerate", "hf-transfer", "sentencepiece", "imageio[ffmpeg]")
     .pip_install(
-        "git+https://github.com/genmoai/models.git@075b6e36db58f1242921deff83a1066887b9c9e1"
-    )
+        "git+https://github.com/huggingface/diffusers",
```
charlesfrye, line 45: Why do we need to install diffusers from `git`? Did they not cut a release with Mochi yet? If so, we need to pin the revision, as we did with `genmoai/models`, to avoid surprises.

We also want to do our `pip_install`s all at once when installing ML libraries -- `pip` doesn't check package compat across install command invocations and will happily mess up your environment.

Regardless, we want to include that install instead

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
-    model_revision: str = "8e9673c5349979457e515fddd38911df6b4ca07f",
-):
-    from huggingface_hub import snapshot_download
-    from transformers import T5EncoderModel, T5Tokenizer
+MINUTES = 60
+HOURS = 60 * MINUTES
 
-    model.reload()
-    print("🍡 downloading Mochi model")
+# ## Setting up our Mochi class
```
charlesfrye, line 61: Put a space after headers and after paragraphs -- renders the same on the website and also gives a bit of breathing room in the code.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+# We'll use the `modal.cls` decorator to define a [Modal Class](https://modal.com/docs/guide/lifecycle-functions)
+# which we use to control the lifecycle of our cloud container.
+#
```
charlesfrye, line 64: See comment on line 61 above.

06_gpu_and_ml/text-to-video/mochi.py
charlesfrye: I.e. delete the `#` in L62, but leave the `\n`.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
-):
-    from huggingface_hub import snapshot_download
-    from transformers import T5EncoderModel, T5Tokenizer
+MINUTES = 60
+HOURS = 60 * MINUTES
 
-    model.reload()
-    print("🍡 downloading Mochi model")
+# ## Setting up our Mochi class
+# We'll use the `modal.cls` decorator to define a [Modal Class](https://modal.com/docs/guide/lifecycle-functions)
```
charlesfrye, line 62: Gotcha: it's `app.cls`, not `modal.cls`. I tend to use `@cls` and elide the namespace.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
-
-    T5Tokenizer.from_pretrained("google/t5-v1_1-xxl", legacy=False)
-    T5EncoderModel.from_pretrained("google/t5-v1_1-xxl")
-
-    model.commit()
-    print("🍡 text encoder downloaded")
+class Mochi:
+    # We can stack the `modal.build` and `modal.enter` decorators
+    # because we use the diffusers library, which caches the model to disk on its first run.
+    # This builds our model weights into the container image, ready for future use.
```
charlesfrye, line 83: We actually want to move away from this pattern, c.f. the updated guide [here](https://modal.com/docs/guide/model-weights).

In particular, we want to stop stacking `build` and `enter` most of the time, because it is confusing to users and adds extra weight to the build step (e.g. the `self.pipe` method calls on L92-3).

charlesfrye: We'd prefer to add the weights to a `Volume` instead. We don't have a great pattern for that with a pure `modal deploy`, but in this case we can just add it to the `local_entrypoint`. Use a separate Function to do the download. You can re-use the code in the Mochi `cls` if you want to still take advantage of HF's auto-loading/caching.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+        
+        # save to disk using prompt as filename
+        for char in string.punctuation:
+            prompt = prompt.replace(char, "")
+        prompt = prompt.replace(" ", "_")
+        prompt = prompt[:230] # since filenames can't be longer than 255 characters
+        mp4_name = str(int(time.time())) + "_" + prompt + ".mp4"
+        
+        export_to_video(frames, Path(OUTPUTS_PATH) /  mp4_name)
+        outputs.commit()
```
charlesfrye, line 119: This shouldn't be necessary given that you're returning immediately afterwards

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+        negative_prompt=str(negative_prompt),
+        num_inference_steps=int(num_inference_steps), 
+        guidance_scale=float(guidance_scale),
+        num_frames=int(num_frames),
+    )
+    print("🍡 video saved to volume at "+mp4_name)
+
     local_dir = Path("/tmp/mochi")
     local_dir.mkdir(exist_ok=True, parents=True)
-    download_model.remote()
```
charlesfrye, line 188: Looks like the original code had the model downloaded to a Volume in the `local_entrypoint`, cf comments on L83 above.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+        frames = self.pipe(
+            prompt=prompt, 
+            negative_prompt=negative_prompt,
+            num_inference_steps=num_inference_steps, 
+            guidance_scale=guidance_scale,
+            num_frames=num_frames,
+        ).frames[0]
+        
+        # save to disk using prompt as filename
+        for char in string.punctuation:
```
charlesfrye, line 112: Let's put all this logic in a separate function, `slugify`, at the end of the file. Otherwise it's a bit distracting.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
 
 # ## Running Mochi inference
 
 # We can trigger Mochi inference from our local machine by running the code in
 # the local entrypoint below.
 
-# It ensures the model is downloaded to a remote volume,
-# spins up a new replica to generate a video, also saved remotely,
-# and then downloads the video to the local machine.
+# It ensures the model is downloaded into the image,
```
charlesfrye, line 127: Nit: I wouldn't include this language if the model is in the image. It's not the `local_entrypoint`'s responsibility to build an image. This language was present because of the call to `download_model.remote`.

Assuming we add back in the model download call, we'd revert to the original text.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+# and then downloads that video to the local machine.
 
 # You can trigger it with:
 
 # ```bash
 # modal run --detach mochi
 # ```
-
+# 
+# Optional command line flags are:
```
charlesfrye, line 137: We should just direct them to the `--help` instead.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+# --prompt="your prompt"
```
charlesfrye, line 138: This didn't render well at all, FYI:
Screenshot 2024-11-19 at 3 15 40 PM

You'd want to put this in triple-backticks and add syntax highlighting.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
+    
     mochi = Mochi()
+    mp4_name = mochi.generate.remote(
+        prompt=str(prompt), 
+        negative_prompt=str(negative_prompt),
+        num_inference_steps=int(num_inference_steps), 
+        guidance_scale=float(guidance_scale),
+        num_frames=int(num_frames),
+    )
+    print("🍡 video saved to volume at "+mp4_name)
```
charlesfrye, line 166: cleaner:

```suggestion
    print("🍡 video saved to volume at", mp4_name)
```

I tend to prefer f-strings, since they are equally, if not more, readable and are more extensible.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
-            json_path = os.path.splitext(output_path)[0] + ".json"
-            with open(json_path, "w") as f:
-                json.dump(args, f, indent=4)
-
-        outputs.commit()
-        print(f"Video saved remotely at: {output_path}")
-        return output_path
-
-
-# ## Addenda
```
charlesfrye, line 338: Bring this back if you add `slugify` as suggested above.

06_gpu_and_ml/text-to-video/mochi.py
```diff
@@ -28,338 +27,147 @@
...
 
-image = (
-    modal.Image.from_registry(f"nvidia/cuda:{tag}", add_python="3.11")
-    .entrypoint([])
-    .apt_install("git", "ffmpeg")
-    .pip_install("torch==2.4.0", "packaging", "ninja", "wheel", "setuptools")
-    .pip_install("flash-attn==2.6.3", extra_options="--no-build-isolation")
+image = (modal.Image.debian_slim()
+    .apt_install("git")
+    .pip_install("torch", "accelerate", "hf-transfer", "sentencepiece", "imageio[ffmpeg]")
```
charlesfrye, line 43: We want to pin versions here. The PR template indicates the policy. Generally, we pin to precise versions pre 1.0 to `x.y` for post 1.0. But with ML libraries, I am sometimes stricter, e.g. pinning `transformers` to a specific version.

charlesfrye (changes requested):

06_gpu_and_ml/text-to-video/mochi.py
erik-dunteman: Before I added `.commit()`, I'd gotten a File Not Exists error from the local entrypoint when immediately downloading from the volume after function return. If a volume isn't fully committed on function return (without manually calling .commit), is that a modal bug?

charlesfrye: Oh, actually that's right -- we don't guarantee that. Always trips me up because I expect it to work that way.

06_gpu_and_ml/text-to-video/mochi.py
erik-dunteman: no release cut yet! pinning revision (both model and git hash)

06_gpu_and_ml/text-to-video/mochi.py
erik-dunteman: You'll note in upcoming change that I've moved to a download_model function, but I won't call it from the local entrypoint. I'll rely on the users calling the function directly if they want a cached model, else it'll download to volume on the from_pretrained on their first @enter

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-55fe7fa.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9828910.modal.run

06_gpu_and_ml/text-to-video/mochi.py
charlesfrye: Oh interesting idea! I like that pattern.

charlesfrye (approved):

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c985e60.modal.run

PR #988, opened 2024-11-22
brings back sdxl turbo example as image_to_image

charlesfrye:
Deleted this one too hastily -- it was our only image-to-image example. Bringing it back, with some small improvements.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-492612c.modal.run

PR #990, opened 2024-11-22
rename stable_diffusion_cli to text_to_image, minor text fixes

charlesfrye:
This example serves as one of our main text-to-image examples and it's no longer just a CLI, so giving it a better name.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-adb6405.modal.run

PR #992, opened 2024-11-22
stop slandering H100s

charlesfrye:
Messed up my SI prefixes -- peta is _quadrillion_, not trillion.

For the record: the [H100 datasheet](https://resources.nvidia.com/en-us-tensor-core/nvidia-tensor-core-gpu-datasheet) from NVIDIA reports ~4 petaFLOPS "with sparsity". The rule of thumb is to cut in half for the much more typical dense case, cf Stas Bekman's ML Engineering notes.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-0bcd551.modal.run

Issue #861, opened 2024-09-02
ComfyUI cli returns 500 server erorr, TimeoutError: timed out.


florisvanderharst:
While running the comfyapp.py and running the comfyclient.py locally - I'm getting errors while the same workflow works in the UI.

After running with verbose on, we do see that comfy is indeed running.

```
Executing workflow: /root/47c2edb7f48f4d4b9f64bd327951b733.json
Executing : Cascade Resolutions - CascadeResolutions (53)
Executing : Load Checkpoint - CheckpointLoaderSimple (153)
Executing : Lora Loader Stack (rgthree) (186)
Executing : ModelSamplingFlux (55)
Executing : description of the scene - CLIPTextEncodeFlux (89)
Executing : CLIP Text Encode (Prompt) - CLIPTextEncode (35)
Executing : ConditioningZeroOut (56)
Executing : Empty Latent Image - EmptyLatentImage (5)
Executing : KSampler (32)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━             73% 0:00:49  
```

but errors out randomly after.

```
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /usr/local/lib/python3.11/site-packages/websocket/_socket.py:118 in recv     │
│                                                                              │
│   115 │   │   if sock.gettimeout() == 0:                                     │
│   116 │   │   │   bytes_ = sock.recv(bufsize)                                │
│   117 │   │   else:                                                          │
│ ❱ 118 │   │   │   bytes_ = _recv()                                           │
│   119 │   except TimeoutError:                                               │
│   120 │   │   raise WebSocketTimeoutException("Connection timed out")        │
│   121 │   except socket.timeout as e:                                        │
│                                                                              │
│ ╭───────────────────────────────── locals ─────────────────────────────────╮ │
│ │   _recv = <function recv.<locals>._recv at 0x7fc6685dd620>               │ │
│ │ bufsize = 2                                                              │ │
│ │    sock = <socket.socket fd=4, family=2, type=1, proto=6,                │ │
│ │           laddr=('127.0.0.1', 41828), raddr=('127.0.0.1', 8188)>         │ │
│ ╰──────────────────────────────────────────────────────────────────────────╯ │
│                                                                              │
│ /usr/local/lib/python3.11/site-packages/websocket/_socket.py:97 in _recv     │
│                                                                              │
│    94 │                                                                      │
│    95 │   def _recv():                                                       │
[...]

PR #997, opened 2024-11-26
use basic http polling for api

kning:
there have been cases of users intermittently hitting [websocket timeout errors](https://github.com/modal-labs/modal-examples/issues/861) in api mode. other customers i've talked to have mentioned how using plain http requests seems to solve that issue, so this reimplements the api portion of the example to just plain http.

did some basic load testing (up to 30 concurrent requests) and wasn't able to observe any errors in a 10 minute or so time window.

### Type of Change

- [ ] New example
- [ ] Other (changes to the codeb
Screenshot 2024-11-26 at 3 44 58 PM
ase, but not to examples)

## Checklist

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/comfyui/comfyapp.py
+++ b/06_gpu_and_ml/comfyui/comfyapp.py
@@ -197,24 +197,52 @@ class ComfyUI:
         subprocess.run(cmd, shell=True, check=True)
 
     @modal.method()
-    def infer(self, workflow_path: str = "/root/workflow_api.json"):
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-865c904.modal.run

kning:
note: i closed this because the issue of random timeouts was not necessarily fixed by this change; the underlying issue is sometimes the comfyui server dies and so you have to poll it periodically and restart it if it does

PR #1002, opened 2024-12-04
resolve incompatibility issue between Gradio and Pydantic

charlesfrye:
and also pin the Python version


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d44e5b3.modal.run

PR #1003, opened 2024-12-04
removes unused 'tags' tag

charlesfrye:
and removes a few stray `runtimes` tags that we don't need anymore, since all examples are run in `gvisor`.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a7d4295.modal.run

PR #1004, opened 2024-12-04
write whitespace in markdown as blank lines, not empty comments

charlesfrye:
This style looks cleaner to me -- it breaks up the long comments in a way that is visually similar to the effect of the whitespace once the contents are actually rendered.

I've been making this change piecemeal, as I update examples. I've been happy so far and so am editing some of our top examples (e.g. featured examples) to use this format.

### DIFF
--- a/05_scheduling/hackernews_alerts.py
+++ b/05_scheduling/hackernews_alerts.py
@@ -1,13 +1,14 @@
 # ---
 # lambda-test: false
 # ---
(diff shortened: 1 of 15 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-1a23aec.modal.run

PR #1008, opened 2024-12-04
change default modal environment for deployment

charlesfrye:
and also check for the example's perferred environment (via the `env` tag)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3028af9.modal.run

PR #1009, opened 2024-12-04
prioritize metadata env over environment variables

charlesfrye:
That way we can set the default via an environment variable, but over-ride in the frontmatter.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6fb9cd6.modal.run

PR #1016, opened 2024-12-07
Update boltz.py

greenrace666:
I got few errors in previous file , hence , I've updated the following:
- Remove MSA code
- boltz uses --use_msa_server
- code recieves --input
- L4 gpu with max timeout (can change)
- export as zip 
- @charlesfrye

### DIFF
--- a/10_integrations/algolia_indexer.py
+++ b/10_integrations/algolia_indexer.py
@@ -46,6 +46,7 @@ CONFIG = {
     },
     "stop_urls": [
         "https://modal.com/docs/reference/modal.Stub",
+        "https://modal.com/gpu-glossary",
     ],
     "start_urls": [
         {


charlesfrye:
Hey @greenrace666, thanks for the PR!

Could you please explain all of the changes that you have made and your rationale for making them?

For example, you removed the code to parse MSAs from the input YAML. Boltz-1's YAML format includes support for linked MSA files, which this removes. Additionally, you're now always passing in the `--use_msa_server` flag, which previously was left as an option for users to pass if they wanted.

greenrace666:
Sure 
- I removed the MSA code since automatic msa prediction would be very user friendly and quick for an average code user
-U now provide custom path to yaml using --input (again user friendly)
-Prob should change back to H100 , L4 takes longer
-zip can be easily extracted cross system
@charlesfrye

PR #1017, opened 2024-12-07
use MSA server by default in Chai-1

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-061eb69.modal.run

PR #1022, opened 2024-12-18
add example for TRELLIS

charlesfrye:
brings in #1018 by @sandeeppatra96 and @patraxo as a `misc` example

### DIFF
--- a/misc/trellis3d.py
+++ b/misc/trellis3d.py
@@ -1,0 +1,252 @@
+"This example originally contributed by @sandeeppatra96 and @patraxo on GitHub"
+import logging
+import tempfile
+import traceback
+
+import modal
+import requests
+from fastapi import HTTPException, Request, Response, status
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+REPO_URL = "https://github.com/microsoft/TRELLIS.git"
+MODEL_NAME = "JeffreyXiang/TRELLIS-image-large"
+TRELLIS_DIR = "/trellis"
+MINUTES = 60
+HOURS = 60 * MINUTES
+
+cuda_version = "12.2.0"
+flavor = "devel"
+os_version = "ubuntu22.04"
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3dc214a.modal.run

charlesfrye:
Super cool example!

Screenshot 2024-12-18 at 9 36 39 AM

A few things before we can bring in as a main-line example that'd show up on the docs site:
- The build process looks pretty messy still (multiple pip install steps that conflict on versions of e.g. Triton, a variety of CUDA versions e.g. 11.8, 12.2, & 12.4). And do we need all of those env vars, e.g. the compiler flags?
- Pin versions for the `pip_install`ed libraries
- Put the model into a Volume instead of the Image, make sure no models are downloaded after the first `@enter`
- Split out the web serving from the underlying function
- Write up the example in the "Literate Programming" style

sandeeppatra96:
> Super cool example!
> 
> Screenshot 2024-12-18 at 9 36 39 AM
> A few things before we can bring in as a main-line example that'd show up on the docs site:
> 
> * The build process looks pretty messy still (multiple pip install steps that conflict on versions of e.g. Triton, a variety of CUDA versions e.g. 11.8, 12.2, & 12.4). And do we need all of those env vars, e.g. the compiler flags?
> * Pin versions for the `pip_install`ed libraries
> * Put the model into a Volume instead of the Image, make sure no models are downloaded after the first `@enter`
> * Split out the web serving from the underlying function
> * Write up the example in the "Literate Programming" style

I had multiple issues as it was unable to find CUDA_HOME, also even though some libaries were being installed without pin versions they were still not working and even caused issues later on while inferencing. e.g. nvdiffrast and diso library

charlesfrye:
Word! I've tasked [our Devin integration](https://github.com/modal-labs/modal-examples/pull/1023) with figuring out the install process.

PR #1024, opened 2024-12-20
speed up blender with L40S

charlesfrye:
at roughly the same cost (and still under a dollar), run the blender rendering in ~3 minutes instead of ~5 with the new L40S GPUs


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2e3c8f5.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ae516e4.modal.run

PR #1027, opened 2024-12-29
Add discord bot example

advay-modal:
Part of initiative to split out boombot into 2 separate examples, one for discord and one for musicgen.

In this example I make a discord bot that just takes a city as input and returns the weather forecast for the next few days

### Type of Change

- New example

## Checklist

- Example is testable in synthetic monitoring system, or `lambda-test: false` is added to example frontmatter (`---`)
- Example is documented with comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style.
- Example does _not_ require third-party dependencies to be installed locally
- Example pins its dependencies
- Example pins container images to a stable tag, not a dynamic tag like `latest`
- Example specifies a `python_version` for the base image, if it is used
- Example pins all dependencies to at least minor version, `~=x.y.z` or `==x.y`
- Example dependencies with `version < 1` are pinned to patch version, `==0.y.z`

## Outside contributors


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-876e0ff.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f678863.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-df4c7a6.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5fbf2b4.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2a88f11.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f4dc284.modal.run

advay-modal:
Screenshot 2024-12-29 at 5 44 18 PM

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c594693.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-65995a4.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d021a61.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-51ed462.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ea5459b.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-db54a17.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3654188.modal.run

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
+# ---
+# lambda-test: false
+# ---
+
+# # Serve a Discord Bot on Modal
+
+# (quick links: [try it out on Discord](https://discord.gg/nR96BxPu))
```
charlesfrye, line 7: I think it's a good idea to host an example, but we should take care to:
1. Make it a Modal-branded Discord server.
2. Restrict all permissions for members except running slash commands.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d8a4903.modal.run

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
+# ---
+# lambda-test: false
```
charlesfrye, line 2: Why not test this?

Ideally, we QA this two ways:
1. Set `deploy: true` so that the example is deployed continuously and include a link to the hosted version so that users can try it and complain to us when it's broken
1. Use `modal run` in Lambda testing to check that the example is working (either the full deployed version or some subset of the logic).

It's a bit tough to check that the whole example application is working because we are not Discord (the requests have to be signed using Discord's private key and the responses are sent to a URL defined by an interaction token from Discord).

So I would maybe advocate for splitting up the logic here -- the `get_weather_forecast_for_city` function handles talking to the external API AND sending back to Discord. If it just talked to the external API, we could wrap it in another function that handles Discord integration.

Decomposing it also provides an opportunity to explain what's going on in the logic -- why do we have these interaction tokens and application IDs? What's with `session.patch` (that's a rare HTTP verb)? This all comes from the Discord docs, to which we should link.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+
+# 1. Go to the
+#    [Discord Developer Portal](https://discord.com/developers/applications) and
+#    login with your Discord account.
+# 2. On the portal, go to **Applications** and create a new application by
+#    clicking **New Application** in the top right next to your profile picture.
+# 3. [Create a custom Modal secret](https://modal.com/docs/guide/secrets) for your Discord bot.
+#    On Modal's secret creation page, select 'Discord'. Copy your Discord application’s
+#    **Public Key** (in **General Information**) and paste the value of the public key
+#    as the value of the `DISCORD_PUBLIC_KEY` environment variable.
```
charlesfrye, line 33: This is actually pretty weird -- the public key is not secret nor is it really ours. It's owned by Discord and they use it to authenticate with _us_. It is specific to our app, but it's actually public information.

Leaking it is harmless on its own, because Discord is the entity that takes actions using it. An adversary would need to compromise our secrets or Discord's in order to trigger Discord to use the key.

Really, we should put the Secrets used to _register_ the slash command into Modal, along with the logic, and then this public key can come along for the ride.

See [this code](https://github.com/the-full-stack/ask-fsdl/blob/b725c0ecdf0d484b48b79e32c37905614522ba4d/bot.py#L207) I wrote for Full Stack Deep Learning. The whole thing is a lot cleaner than the code in Boombot.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+
+# To connect our model to a Discord bot, we’re first going to create an
+# application on the Discord Developer Portal.
+
+# 1. Go to the
+#    [Discord Developer Portal](https://discord.com/developers/applications) and
+#    login with your Discord account.
+# 2. On the portal, go to **Applications** and create a new application by
+#    clicking **New Application** in the top right next to your profile picture.
+# 3. [Create a custom Modal secret](https://modal.com/docs/guide/secrets) for your Discord bot.
```
charlesfrye, line 30:
```suggestion
# 3. [Create a custom Modal Secret](https://modal.com/docs/guide/secrets) for your Discord bot.
```

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+#    On Modal's secret creation page, select 'Discord'. Copy your Discord application’s
```
charlesfrye, line 31:
```suggestion
#    On Modal's Secret creation page, select 'Discord'. Copy your Discord application’s
```

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+#    **Public Key** (in **General Information**) and paste the value of the public key
+#    as the value of the `DISCORD_PUBLIC_KEY` environment variable.
+#    Name this secret `weather-discord-secret`.
```
charlesfrye, line 34:
```suggestion
#    Name this Secret `weather-discord-secret`.
```

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+# }' "https://discord.com/api/v10/applications/$CLIENT_ID/commands"
+# ```
+
+
+# This will register a Slash Command for your bot named `get_weather`, and has a
+# parameter called `city`. More information about the
+# command structure can be found in the Discord docs
+# [here](https://discord.com/developers/docs/interactions/application-commands).
+
+# ## Defining the asgi app with modal
```
charlesfrye, line 73: There's a lot of code in between this title/description and the actual ASGI app creation.

Even apart from splitting up the logic further, the prose and code organization also needs to be split up further. The logic for what our service does is separate from how we serve it -- even though our service here is pretty simple.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+
+# Let's get the imports out of the way and define an [`App`](https://modal.com/docs/reference/modal.App)
+
+import json
+
+import modal
+
+app = modal.App("example-discord-weather-bot")
+
+# We define an [Image](https://modal.com/docs/guide/images) that has the [`python-weather`](https://github.com/null8626/python-weather) package and
```
charlesfrye, line 88: This doesn't add much beyond what's in the code.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+# the [FastAPI](https://fastapi.tiangolo.com/) package installed.
+
+image = modal.Image.debian_slim(python_version="3.11").pip_install(
+    "python-weather==2.0.7", "fastapi[standard]==0.115.4", "pynacl==1.5.0"
+)
+
+# We define a function that uses the `python_weather` library to get the weather of a city
+# Note that since Discord requires an interaction response within 3 seconds, we
```
charlesfrye, line 96: This prose is in the wrong place -- the function below doesn't have any of this logic.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+        except Error:
+            message = "An error occurred, please check city name"
+
+    json_payload = {"content": message}
+    async with aiohttp.ClientSession() as session:
+        async with session.patch(interaction_url, json=json_payload) as resp:
+            print(await resp.text())
+
+
+# We now define an asgi app using the Modal ASGI syntax [@asgi_app](https://modal.com/docs/guide/webhooks#serving-asgi-and-wsgi-apps).
```
charlesfrye, line 131: This comment doesn't add much.

This is where I would put the information that is beneath the "Defining the asgi app" header above.

I would also comment on the `keep_warm` choice -- the cost and why we're doing it.

Note that the need to keep a container warm to hit Discord's time limits pulls us away from a "true serverless" model. It's also pretty unlikely that someone would need to scale up a Discord bot beyond what a single FastAPI replica can handle, so we should focus on the idea of scaling up the _backend_ and on separation of concerns between the backend service and the API in front of it.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+@app.function(
+    secrets=[modal.Secret.from_name("weather-discord-secret")],
+    keep_warm=1,  # eliminates risk of container startup making discord ack time too long
```
charlesfrye, line 136: We should also `allow_concurrent_inputs` here -- no need to spin up a replica for every call from Discord.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+)
+
+# We define a function that uses the `python_weather` library to get the weather of a city
+# Note that since Discord requires an interaction response within 3 seconds, we
+# use [`spawn`](https://modal.com/docs/reference/modal.Function#spawn) to kick off
+# `get_weather_for_city`as a background task from the asgi app while returning a `defer` message to
+# Discord within the time limit.
+
+
+@app.function(image=image)
```
charlesfrye, line 102: Any part that uses `async` should have `allow_concurrent_inputs` set.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+    image=image,
+)
+@modal.asgi_app()
+def web_app():
+    from fastapi import FastAPI, HTTPException, Request
+    from fastapi.middleware.cors import CORSMiddleware
+
+    web_app = FastAPI()
+
+    # allow CORS
```
charlesfrye, line 146: A comment or at least a link here would be good.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+    web_app.add_middleware(
+        CORSMiddleware,
+        allow_origins=["*"],
+        allow_credentials=True,
+        allow_methods=["*"],
+        allow_headers=["*"],
+    )
+
+    @web_app.post("/get_weather_forecast")
+    async def get_weather_forecast_api(request: Request):
```
charlesfrye, line 156: The names on the functions in this are a bit messy. Could we clean them up? Like this could just be `get_weather_forecast` or just `get_weather`, as could the function above.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+        import os
+
+        from nacl.exceptions import BadSignatureError
+        from nacl.signing import VerifyKey
+
+        # Verify the request using the Discord public key
```
charlesfrye, line 162: Verify the request _is from Discord_

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+        message = timestamp.encode() + body
+
+        try:
+            verify_key.verify(message, bytes.fromhex(signature))
+        except BadSignatureError:
+            raise HTTPException(status_code=401, detail="Invalid request")
+
+        # Parse request
+        data = json.loads(body.decode())
+        if data.get("type") == 1:  # ack ping from Discord
```
charlesfrye, line 179: An enum would be good here, see [this code](https://github.com/the-full-stack/ask-fsdl/blob/b725c0ecdf0d484b48b79e32c37905614522ba4d/bot.py#L23-L34).

07_web_endpoints/discord_bot.py
charlesfrye: Ah, we can't shadow it as-is because we later need a handle to call `.spawn` on within this scope.

I think we might be able to remove that constraint if we change the composition here -- but it might just end up uglier. In that case, can only get slightly cleaner names.

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+# We now define an asgi app using the Modal ASGI syntax [@asgi_app](https://modal.com/docs/guide/webhooks#serving-asgi-and-wsgi-apps).
+
+
+@app.function(
+    secrets=[modal.Secret.from_name("weather-discord-secret")],
+    keep_warm=1,  # eliminates risk of container startup making discord ack time too long
+    image=image,
+)
+@modal.asgi_app()
+def web_app():
```
charlesfrye, line 140: I've started to dislike `web_app` as a name. Really, it's an API! So we should call these functions `api` or `web_api` in our examples. That also makes the final URL shorter.

charlesfrye (changes requested):

07_web_endpoints/discord_bot.py
```diff
@@ -0,0 +1,230 @@
...
+# We define a function that uses the `python_weather` library to get the weather of a city
+# Note that since Discord requires an interaction response within 3 seconds, we
+# use [`spawn`](https://modal.com/docs/reference/modal.Function#spawn) to kick off
+# `get_weather_for_city`as a background task from the asgi app while returning a `defer` message to
+# Discord within the time limit.
+
+
+@app.function(image=image)
+async def get_weather_forecast_for_city(
+    city: str, interaction_token, application_id
```
charlesfrye, line 104: This function is doing a bit too much -- figuring out the Discord URL to talk to, hitting a different external API, and then submitting the request. That makes it harder to write about and harder to effectively name.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-fcf035c.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-1834262.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a90ca64.modal.run

07_web_endpoints/discord_bot.py
advay-modal: What's the advantage of having the ability to register a slash command as a modal function? Imo if we do that we'd have to get the user to run this separate function before they run anything else on the app

07_web_endpoints/discord_bot.py
advay-modal: "so we should focus on the idea of scaling up the backend and on separation of concerns between the backend service and the API in front of it." -> wdym by focus?

07_web_endpoints/discord_bot.py
charlesfrye: 1. Creating a slash command in Modal requires secret information, which you can put in the Secret. It's a better demonstration of what Secrets are for. It's a bit weird that we're using it for configuration alone.
2. Including the setup in the code itself makes the example more self-contained and less error-prone (are you confident that snippet runs in _all_ shells?). You can add a call to that Function in the local entrypoint, you just need to make it idempotent (e.g. with a `force` or `overwrite` flag, as in the linked code [here](https://github.com/the-full-stack/ask-fsdl/blob/b725c0ecdf0d484b48b79e32c37905614522ba4d/bot.py#L235-L253)).

07_web_endpoints/discord_bot.py
charlesfrye: Yeah, "focus on" as in "focus the prose in the example on".

My point is that the Discord bot hosting itself is not super compelling, especially now that we have a floor for CPU usage.

charlesfrye (approved):

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-997f8ca.modal.run

PR #1034, opened 2025-01-03
fix librispeech download

charlesfrye:
Dependency issue. Bumping everything.

Also some noisy formatting changes while I'm here.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-99875d6.modal.run

PR #1035, opened 2025-01-03
allow control of blender render from local client

charlesfrye:
Setting the frame_skip parameter >1 during monitoring should also resolve the crashes that didn't reproduce elsewhere.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-589aa8a.modal.run

PR #1039, opened 2025-01-09
Update trtllm_llama.py

KarthikeyanVijayanEasyLLM:
I ran into following exception when trying out the example. Fixing that with latest CUDA image and latest TensorRT-LLM package.

Error:
Step 2: RUN python /root/convert.py --model_dir=/root/model/model_input --output_dir=/root/model/model_ckpt --tp_size=1 --dtype=float16 --qformat=fp8 --kv_cache_dtype=fp8 --calib_size=512
AttributeError: module 'pynvml' has no attribute '__version__'
Terminating task due to error: failed to run builder command "python /root/convert.py --model_dir=/root/model/model_input --output_dir=/root/model/model_ckpt --tp_size=1 --dtype=float16 --qformat=fp8 --kv_cache_dtype=fp8 --calib_size=512": container exit status: 1
Runner failed with exception: task exited with failure, status = exit status: 1
Stopping app - uncaught exception raised locally: RemoteError('Image build for im-Pf6HdxV8HNda62wDKk62ih failed with the exception:\ntask exited with failure, status = exit status: 1').

### Type of Change

- [ ] New example

## Checklist

### DIFF
--- a/06_gpu_and_ml/llm-serving/trtllm_llama.py
+++ b/06_gpu_and_ml/llm-serving/trtllm_llama.py
@@ -57,7 +57,7 @@ import modal
 import pydantic  # for typing, used later
 
 tensorrt_image = modal.Image.from_registry(
-    "nvidia/cuda:12.4.1-devel-ubuntu22.04",
+    "nvidia/cuda:12.5.1-devel-ubuntu22.04",
     add_python="3.10",  # TRT-LLM requires Python 3.10
 ).entrypoint([])  # remove verbose logging by base image on entry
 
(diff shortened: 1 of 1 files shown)


charlesfrye:
Thanks for reporting!

I did some digging and it looks like the issue arises from a dependency, `pynvml`, so we can resolve by just pinning that dependency's version more tightly. Applied that fix in #1040.

Technically, the 12.5 version of the CUDA Driver/Runtime APIs is not supported by the underlying host driver version we install (550.90.07) and so solving the issue by updating those libraries might lead to other issues.

PR #1040, opened 2025-01-10
pin pynvml version to avoid breaking change to version API

charlesfrye:
resolves the issue raised in #1039


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6ecea99.modal.run

PR #1041, opened 2025-01-11
update the OCR example

charlesfrye:
Building on the work by @vitorialima to update this example:

- Updates the model from DONUT to GOT-OCR 2.0
- Updates dependencies -- DONUT was quite old and capped versions of other components, e.g. Python<3.10
- Stores the model in a Volume instead of an Image
- Spruces up the frontend to match our design style

Screenshot 2025-01-10 at 4 51 56 PM


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-81e2453.modal.run

PR #1042, opened 2025-01-11
update esm3 to new 'mount' style

charlesfrye:
as picked up by synthetic monitoring, we are deprecating the Mount style and will instead use image methods with optional copies


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-59157bc.modal.run

shariqm-modal:
sg, thanks

erikbern:
nice, feels cleaner

PR #1043, opened 2025-01-11
add section on proxy authentication

charlesfrye:
This PR adds a section on the proxy auth feature to the `basic_web` example.

I tried pretty hard to improve on the sample shell command from the proxy auth guide but came up empty -- no combination of curl/wget flags or netrc settings avoided the need to base64 encode by hand.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6a248e1.modal.run

PR #1045, opened 2025-01-15
prose edits, lambda-test compat, and an extra argument

charlesfrye:
Most of the prose changes should be uncontroversial, but a quick technical gut check would be appreciated.

I added an argument to allow users to pass arguments to `torchrun`. Less about that being super useful and more about showing the pattern (which could be used to pass arguments to a script) and showing a bit of Modal's extra flexibility over other platforms.

The other code changes are cosmetic.

### DIFF
--- a/14_clusters/simple_torch_cluster_script.py
+++ b/14_clusters/simple_torch_cluster_script.py
@@ -18,7 +18,7 @@ MASTER_RANK = 0
 
 # This `run` function performs a simple distributed data transfer between containers
 # using the specified distributed communication backend.
-#
+
 # An example topology of the cluster when WORLD_SIZE=4 is shown below:
 #
 #        +---------+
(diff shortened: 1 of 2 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a0b7005.modal.run

14_clusters/simple_torch_cluster_script.py
```diff
@@ -33,10 +33,10 @@
 # +--+--+  +--+--+  +--+--+
 # |Rank 1| |Rank 2| |Rank 3|
 # +-----+  +-----+  +-----+
-#
```
thundergolfer, line 36: What's wrong with these?

charlesfrye: Oh this is pretty minor, but: I like to put blank lines in each spot where there would be whitespace in a rendered document. It's a bit closer to the look of the rendered doc and (at least to me) suggests that this document isn't purely intended as source code.

Minor enough that I wouldn't've done it if I weren't already making changes.

PR #1046, opened 2025-01-15
make clearer

kning:
- Mount hf_hub_download cache dir to volume (same pattern as [run flux fast](https://modal.com/docs/examples/flux))
- Generally slim down a lot of copy

### Type of Change

- [ ] New example

## Checklist

## Outside contributors


06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -72,126 +48,105 @@
     .run_commands(  # use comfy-cli to install the ComfyUI repo and its dependencies
         "comfy --skip-prompt install --nvidia"
```
kning, line 49: minor note: i don't think there's a way to pin which version of comfyui you're installing, which represents some risk

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-7e1f16b.modal.run

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -3,46 +3,32 @@
...
-#
-# To serve the workflow in this example as an API:
-# 1. Download the Flux models to a Modal [Volume](/docs/guide/volumes):
-# ```bash
-# modal run 06_gpu_and_ml/comfyui/comfyapp.py::download_models
-# ```
-#
-# 2. Stand up the ComfyUI server in development mode:
+# To run this simple text-to-image [Flux Schnell workflow](https://github.com/modal-labs/modal-examples/blob/main/06_gpu_and_ml/comfyui/workflow_api.json) as an API:
+# 1. Stand up the ComfyUI server in development mode:
```
kramstrom, line 13:
```suggestion
# 1. Start up the ComfyUI server in development mode:
```

06_gpu_and_ml/comfyui/comfyapp.py
kramstrom: Ah I was under the impression the comfy version was pinned by the comfy-cli version but maybe that's not the case

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -72,126 +48,105 @@
     .run_commands(  # use comfy-cli to install the ComfyUI repo and its dependencies
         "comfy --skip-prompt install --nvidia"
     )
-    .add_local_file(
+    .add_local_file(  # copy ComfyUI workflow JSON to the container
         Path(__file__).parent / "workflow_api.json",
         "/root/workflow_api.json",
         copy=True,
```
kramstrom, line 54: We don't need `copy=True` here if we move this step to the end of image build since we don't use it in the subsequent steps. If we remove `copy=True` and set it as the last step of the image build we will avoid rebuilding subsequent image layers when we modify the workfow file

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -72,126 +48,105 @@
...
-#
-# First we need to run a function to download the Flux models to a Modal Volume.
+    flux_model = hf_hub_download(
+        repo_id="Comfy-Org/flux1-schnell",
+        filename="flux1-schnell-fp8.safetensors",
+        cache_dir="/cache",
+    )
 
-vol = modal.Volume.from_name("comfyui-models", create_if_missing=True)
+    # symlink the model to the right ComfyUI directory
```
kramstrom, line 87: Hm, why are we symlinking in the `hf_download` function that is run as an image layer? Shouldn't we only need so symlink it in the `@enter` method of the function?

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -72,126 +48,105 @@
...
+# ## Running ComfyUI as an API
 #
-# Here's the basic breakdown of how we do it:
-# 1. We stand up a "headless" ComfyUI server in the background when the app starts.
-# 2. We define an `infer` method that takes in a workflow path and runs the workflow on the ComfyUI server.
-# 3. We stand up an `api` with `web_endpoint`, so that we can run our workflows as a service.
+# To run a workflow as an API:
+# 1. Stand up a "headless" ComfyUI server in the background when the app starts.
+# 2. Define an `infer` method that takes in a workflow path and runs the workflow on the ComfyUI server.
+# 3. Stand up a web handler `api` with `web_endpoint`, so that we can run our workflow as a service and accept inputs from clients.
```
kramstrom, line 137:
```suggestion
# 3. Start up a web handler `api` with `web_endpoint`, so that we can run our workflow as a service and accept inputs from clients.
```

kning:
great feedback @kramstrom working on it

06_gpu_and_ml/comfyui/comfyapp.py
kning: i did it in the `hf_download` function because `hf_hub_download` returns the filepath where the model is saved so wanted to do the symlink within the same scope if that makes sense?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b3fb736.modal.run

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -134,14 +129,20 @@ def ui():
...
     allow_concurrent_inputs=10,  # allow 10 concurrent API calls
     container_idle_timeout=300,  # 5 minute container keep alive after it processes an input; increasing this value is a great way to reduce ComfyUI cold start times
     gpu="L40S",
     volumes={"/cache": vol},
+    mounts=[  # mount workflow JSON that we want to serve to the container
+        modal.Mount.from_local_file(
+            Path(__file__).parent / "workflow_api.json",
+            remote_path="/root/workflow_api.json",
+        )
+    ],
```
kramstrom, line 145: Sorry this was my bad, mounts on classes / functions is about to be deprecated (I think there's already a warning about it or about to be) so the previous implementation was correct with `add_local_file` on the Image. I was just suggesting to put it as the very last step of the image build so we don't need the `copy=True` and subsequent steps aren't invalidated whenever the workflow file changes

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9d64a08.modal.run

kramstrom (approved):
great!

PR #1047, opened 2025-01-16
remove reference to CPU-only clusters

charlesfrye:
cc @thecodingwizard


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-42b5bee.modal.run

PR #1048, opened 2025-01-16
add barrier before process group cleanup

charlesfrye:
iiuc not necessary because our ops are all sync but good to include in case users take this code and run more complex scripts with it

cc @thecodingwizard


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-70d8e44.modal.run

PR #1049, opened 2025-01-16
Update whisper fine-tune example

aksh-at:
Quick update to get training and end-to-end check working. I removed the CLI parsing and forwarding since having a local version of `transformers` was causing issues, will think about it.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-79091e7.modal.run

charlesfrye:
Thanks! Yeah we might want to give this one a bit of love and add it to monitoring.

PR #1050, opened 2025-01-20
[wip] deepseek r1 distilled llama 70B

pawalt:
Keeping this as a WIP since we probably want to keep the Llama example. Only changes are:

- vLLM version
- Number of GPUs (full precision 70B requires 2xH100)
- Chat format argument required in new vLLM version

### DIFF
--- a/06_gpu_and_ml/llm-serving/download_llama.py
+++ b/06_gpu_and_ml/llm-serving/download_llama.py
@@ -1,12 +1,13 @@
 # ---
 # args: ["--force-download"]
 # ---
+
 import modal
 
 MODELS_DIR = "/llamas"
 
-DEFAULT_NAME = "neuralmagic/Meta-Llama-3.1-8B-Instruct-quantized.w4a16"
-DEFAULT_REVISION = "a7c09948d9a632c2c840722f519672cd94af885d"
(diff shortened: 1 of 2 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-dcedce0.modal.run

charlesfrye:
nice! I think we'll keep the smaller model in the main example, since the download time and cold boots for a 70B model are pretty painful.

salman1993:
@pawalt have you looked into enabling tool calling by any chance? 
https://github.com/vllm-project/vllm/blob/6d0e3d372446cde48b387d4d3530e25fc6e06320/vllm/entrypoints/openai/serving_chat.py#L47

i get back responses from the API when i add in `enable_auto_tools` and `tool_parser` params, but tool calling doesn't work that well. my guess is the `granite` parser doesn't work well for deepseek models - we'd need the tool template it was trained with.
```
api_server.chat = lambda s: OpenAIServingChat(
        engine,
        ...
        chat_template_content_format="raw",
        enable_auto_tools=True,
        tool_parser="granite",
    )
```

charlesfrye:
closing since we now have a DeepSeek-R1 example (cf #1057)

PR #1051, opened 2025-01-22
switch to Free Public API API

charlesfrye:
The Bored API went down basically immediately after I published this example and has been hard down for a few days.

In searching for a more stable free public API, I uncovered the [🤯 Free Public API API](https://www.freepublicapis.com/api), which works well for our purposes.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3226d2b.modal.run

PR #1059, opened 2025-01-28
switch model in sample cmd to match change in defaults

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6c6df52.modal.run

PR #1060, opened 2025-01-28
Migrate object lookups in example scripts

mwaskom:
We're deprecating `.lookup` methods on Modal objects to streamline object creation. `.from_name` should now be the One Way To Do It. I have updated the examples accordingly. There were a couple cases where the examples demonstrated a pattern that couldn't be solved with a simple name substitution and required explicitly hydrating the object.

Part of CLI-96

### Type of Change

- [ ] New example


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-70c03d7.modal.run

charlesfrye:
thanks!

PR #1061, opened 2025-01-29
Fix Huggingface Cache environment variable

steve-marmalade:
While testing one of the examples, I noticed that the local huggingface cache was not being used. While reviewing the code, I noticed a typo where the environment variables `HF_HUB_CACHE_DIR` was being set but, [according to the docs](https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache#understand-caching), the correct environment variable is `HF_HUB_CACHE`.

Figured I'd send along a PR so that it helps others.

### Type of Change

- [ ] New example

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/sam/segment_anything.py
+++ b/06_gpu_and_ml/sam/segment_anything.py
@@ -62,7 +62,7 @@ cache_dir = "/cache"
 
 
 @app.cls(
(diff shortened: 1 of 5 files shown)


charlesfrye:
superceded by #1069, which makes the same changes but is inside this repo instead of a fork and so benefits from testing

PR #1068, opened 2025-02-09
switch from shlex to varargs

charlesfrye:
it's cleaner!


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ac289b9.modal.run

PR #1071, opened 2025-02-11
Fix/deepseek openai server

LeTriet17:
---

✍️ **Update GPU type and context size in `deepseek_openai_server.py`**

- Changed GPU type from **A10G** to **L40S** for inference to leverage better performance.
- Reduced `n_ctx` from **32768** to **8096** to improve stability during experiments.
- All other settings remain unchanged.

---

### Type of Change

- [ ] New example  


charlesfrye:
thanks!

PR #1072, opened 2025-02-11
update the vllm oai client script to accept/detect modal environments

charlesfrye:
We construct the web url using information about the vLLM deployment provided by the user or their Modal configuration, but we weren't looking at environments. That lead to a surprising error for @shariqm-modal when his client hit the deployment in the `main` environment of the `modal-labs` workspace, which had been deployed with proxy authorization.

I think there might be a better way to figure out URLs with the `.web_url` attribute of `Function`, but I'm not 100% sure how to handle `serve`, so pushing this fix instead.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-03c6ec1.modal.run

PR #1073, opened 2025-02-12
update whisper pod transcript for re-deploy

charlesfrye:
The Whisper pod-transcriber app went down because it was running an old client.

This PR makes the changes needed to re-deploy. A few URLs are changed because we now deploy examples into the `examples` environment. Numpy released a new major version in the interim, so we pin below it.

Also, the container counting logic wan't working, so I dropped it.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b15bb92.modal.run

PR #1079, opened 2025-02-21
codemod force build

charlesfrye:
This PR reworks the way we trigger forced builds of examples in CI so that we don't cause base images like `debian_slim` to be rebuilt in the `modal-labs` workspace.

It can be tested by re-enabling [the force-build-example workflow](https://github.com/modal-labs/modal-examples/actions/workflows/force-build-example.yml) and triggering a manual workflow dispatch from this branch.

The new `internal/add_force_build.py` script pulls out YAML frontmatter, parses the rest of the file into an AST, and then looks for method names that match certain `modal.Image` methods and appear to be `modal.Image` attribute accesses and adds `force_build=True`.

This is admittedly a hack. It seemed like being certain a method was on a `modal.Image` was going to require full Python type-checking (and `mypy` doesn't seem designed for code modification), but there might be a creative alternative I missed (like something cursed with `inspect`).

I'm willing to scrap this PR and go back to the drawing board if it ends up more trouble than it is worth.

### Type of Change

### DIFF
--- a/.github/workflows/force-build-example.yml
+++ b/.github/workflows/force-build-example.yml
@@ -24,6 +24,10 @@ jobs:
           fetch-depth: 1
       - uses: ./.github/actions/setup
 
-      - name: Run a random example with MODAL_FORCE_BUILD set
+      - name: Add force build to all Image builds
         run: |
-          MODAL_FORCE_BUILD=1 python3 -m internal.run_example
+          find . -name "*.py" | LOGLEVEL=DEBUG xargs python3 internal/add_force_build.py
+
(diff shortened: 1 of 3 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-bdcf328.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-1a2ed4e.modal.run

charlesfrye:
superceded by #1085

PR #1081, opened 2025-02-24
remove unneeded frontmatter and update formatting

charlesfrye:
small cleanup tasks on this example.

side note: it's been failing in monitoring for the last few days due to runs not pulling any prices for any tickers, but i wasn't able to reproduce. there was a build the first time i ran manually, which is sus because i don't see any attempted builds in the monitoring runs.

### DIFF
--- a/03_scaling_out/fetch_stock_prices.py
+++ b/03_scaling_out/fetch_stock_prices.py
@@ -1,29 +1,28 @@
 # ---
 # output-directory: "/tmp/"
-# runtimes: ["runc", "gvisor"]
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-7c60e02.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-76b8093.modal.run

PR #1082, opened 2025-02-26
Update algolia_indexer.py to omit changelog

RohanArepally:
Omitting the change log entirely since it's all on one page and unclear how useful it is to users trying to search for something. For example if something matches in the changelog from last year is that even useful? 

Since it's all on one page users can just ctrl-f on that page to find something.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-940fdb3.modal.run

RohanArepally:
@prbot approve

PR #1086, opened 2025-03-04
Migrate modal.web_endpoint to modal.fastapi_endpoint

mwaskom:
Updates the examples to be 1.0 compatible by substituting `modal.fastapi_endpoint` for `modal.web_endpoint`.

Needs to be merged before https://github.com/modal-labs/modal/pull/20457 as it fixes a cross-reference that needs to change.

Part of CLI-339

### Type of Change

- [ ] New example

### DIFF
--- a/06_gpu_and_ml/comfyui/comfyapp.py
+++ b/06_gpu_and_ml/comfyui/comfyapp.py
@@ -101,9 +101,7 @@ image = (
 )
 
 # Lastly, we copy the ComfyUI workflow JSON to the container.
-image = image.add_local_file(
-    Path(__file__).parent / "workflow_api.json", "/root/workflow_api.json"
-)
+image = image.add_local_file(Path(__file__).parent / "workflow_api.json", "/root/workflow_api.json")
 
 # ## Running ComfyUI interactively
 #
@@ -133,7 +131,7 @@ def ui():
 # To run a workflow as an API:
 # 1. Stand up a "headless" ComfyUI server in the background when the app starts.
 # 2. Define an `infer` method that takes in a workflow path and runs the workflow on the ComfyUI server.
-# 3. Create a web handler `api` with `web_endpoint`, so that we can run our workflow as a service and accept inputs from clients.
+# 3. Create a web handler `api` as a web endpoint, so that we can run our workflow as a service and accept inputs from clients.
 #
 # Group all these steps into a single Modal `cls` object, which we'll call `ComfyUI`.
 @app.cls(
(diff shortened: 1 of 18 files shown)


06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -160,24 +158,20 @@ def infer(self, workflow_path: str = "/root/workflow_api.json"):
 
         # looks up the name of the output image file based on the workflow
         workflow = json.loads(Path(workflow_path).read_text())
-        file_prefix = [
-            node.get("inputs")
-            for node in workflow.values()
-            if node.get("class_type") == "SaveImage"
-        ][0]["filename_prefix"]
+        file_prefix = [node.get("inputs") for node in workflow.values() if node.get("class_type") == "SaveImage"][0][
```
mwaskom, line 161: wtf ruff

06_gpu_and_ml/comfyui/comfyapp.py
```diff
@@ -101,9 +101,7 @@ def hf_download():
 )
 
 # Lastly, we copy the ComfyUI workflow JSON to the container.
-image = image.add_local_file(
-    Path(__file__).parent / "workflow_api.json", "/root/workflow_api.json"
-)
```
mwaskom, line 106: I think this was the first time a lot of examples were exposed to the upgraded `ruff`; sorry for the noise.

charlesfrye (approved):
np on the reformatting. probably should have done a pass back when we upgraded

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8f491d2.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-7306577.modal.run

10_integrations/algolia_indexer.py
```diff
@@ -112,7 +112,7 @@
 # ## The actual function
 #
 # We want to trigger our crawler from our CI/CD pipeline, so we're serving it as a
-# [web endpoint](/docs/guide/webhooks#web_endpoint) that can be triggered by a `GET` request during deploy.
+# [web endpoint](/docs/guide/webhooks) that can be triggered by a `GET` request during deploy.
```
charlesfrye, line 115: We should change this to an absolute URL. That's the style we use elsewhere because it works better when the examples are viewed in editors. It means that we can't catch 404s as easily, but it also reduces what are IMO noisy build errors like this one from a changed fragment identifier.

mwaskom: That makes sense. I see a few others so I'll leave it to be done in all one coherent PR.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-0e04630.modal.run

PR #1089, opened 2025-03-06
make relative modal.com links global, standardize formatting

charlesfrye:
This PR makes all relative URLs, which are implicitly to https://modal.com, explicitly global URLs. That makes the URLs easier to copy from an editor at the expense of making it harder to catch 404s.

It also standardizes some formatting choices, e.g. using blank lines in places where the rendered page will have vertical spacing, like between blocks and around headers.

### DIFF
--- a/01_getting_started/hello_world.py
+++ b/01_getting_started/hello_world.py
@@ -1,15 +1,15 @@
 # # Hello, world!
-#
+
 # This tutorial demonstrates some core features of Modal:
-#
+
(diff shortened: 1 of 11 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-04991f4.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8164757.modal.run

PR #1092, opened 2025-03-06
pin python version in more examples

charlesfrye:
I've left the Python version unpinned in a few examples that we actually test, since they are cases where we'd want the feature to work across Python versions, e.g. the flask/FastAPI serving examples. These examples also generally have the library versions unpinned for the same reason. They're more like canaries for library compatibility issues than are the other examples, which are closer to applications and so have tighter version pinning.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-084f4e0.modal.run

PR #1094, opened 2025-03-10
surface failures from run_random_example

charlesfrye:
The [build-and-run-example workflow](https://github.com/modal-labs/modal-examples/actions/workflows/build-and-run-example.yml) was failing silently because we always returned 0 instead of the example's actual returncode. This PR returns the returncode.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-932ec8f.modal.run

PR #1096, opened 2025-03-13
better defaults for phi, always pre-download

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ef5ecf0.modal.run

Issue #1101, opened 2025-03-17
Using a compromised tj-actions/changed-files GitHub Action


shubham-stepsecurity:
Filing a public issue instead of reporting this as a private vulnerability, as I could not find a security.md file. Moreover, this malware is a publicly known and an urgent issue.

This repo uses a compromised version of tj-actions/changed-files. The compromised action leaks secrets the runner has in memory.

https://github.com/modal-labs/modal-examples/blob/4956aa537f6265539f8b0e0fb122f924a09bd567/.github/workflows/run-examples.yml#L43

This run ids has creds leaked. Please rotate (if applicable) and delete the workflow run.
https://github.com/modal-labs/modal-examples/actions/runs/13865022099/job/38802114787#step:3:63

You can also use https://github.com/step-security/changed-files going forward.

Reference about this incident: https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

charlesfrye:
Thanks for reporting! We are aware and have remediated, cd #1100.

PR #1102, opened 2025-03-17
remove run-examples workflow temporarily

charlesfrye:
Removes the `run-examples` workflow to make clear that we aren't in fact still running the compromised `tj-actions/changed-files` Action, cf #1101.


charlesfrye:
For the record, evidence that we blocked the compromised Action is [here](https://github.com/modal-labs/modal-examples/actions/runs/13904611718).

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-be2b1d5.modal.run

PR #1111, opened 2025-03-28
runs changed examples again

charlesfrye:
The changed-files action was compromised, so we brought the logic in-house.

Tested with `nektos/act` locally, so hopefully this works without needing to push a bunch of commits onto GitHub Actions for testing.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5d8f383.modal.run

PR #1114, opened 2025-03-28
pin all GitHub Actions by hash

charlesfrye:
Implements [SEC-48](https://linear.app/modal-labs/issue/SEC-48/pin-all-github-actions-that-we-use-to-a-specific-hash) for `modal-examples`.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3b3bfa6.modal.run

PR #1116, opened 2025-04-01
use custom cuda image

kning:
A user running this example reported an issue of crash looping containers with the message:

`ImportError: libnvrtc.so.12: cannot open shared object file: No such file or directory`

When I first ran it, it worked because it leveraged the cached build. However when I triggered a force build I also started getting this error (and now the synthetic monitoring has also just started failing). Something strange might be happening with image build caching?

Anyways I was able to resolve by using a custom cuda image.

### Type of Change

- [ ] New example

## Checklist

## Outside contributors


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6ed0735.modal.run

charlesfrye:
Yeah, this is due to `flashinfer` changing its behavior, and we don't have its version pinned here.

PR #1117, opened 2025-04-02
Basic ComfyUI memory snapshot example for sub-2s cold start time

pingren:
- Bump comfy-cli to 1.3.8, use `--fast-deps` to speed up image building with [uv](https://github.com/astral-sh/uv)
- Introduced a simple ComfyUI custom nodes for memory snapshots.
- Updated ComfyUI app to include methods for launching in the background and restoring from snapshots.

To test memory snapshot, `modal deploy 06_gpu_and_ml/comfyui/comfyapp.py` and start and stop the container for a few times.

image

I've successfully implemented memory snapshot preloading for models, but don't believe the current implementation is ideal for an example. Let's go with this PR as a first step.

### Type of Change

- [ ] New example

## Checklist

## Outside contributors

### DIFF
--- a/06_gpu_and_ml/comfyui/comfyapp.py
+++ b/06_gpu_and_ml/comfyui/comfyapp.py
@@ -14,13 +14,13 @@
 # 1. Start up the ComfyUI server in development mode:
 
 # ```bash
-# modal serve 06_gpu_and_ml/comfyui/comfyapp.py
+# modal deploy 06_gpu_and_ml/comfyui/comfyapp.py
 # ```
 
 # 2. In another terminal, run inference:
 
 # ```bash
-# python 06_gpu_and_ml/comfyui/comfyclient.py --dev --modal-workspace $(modal profile current) --prompt "Surreal dreamscape with floating islands, upside-down waterfalls, and impossible geometric structures, all bathed in a soft, ethereal light"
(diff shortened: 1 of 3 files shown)


kning:
thanks for this, confirming it works for me as well :).

two thoughts:
- i wonder how hard it would be to push your memory_snapshot custom node to core. they seem like pretty small changes that don't change behavior and we all have a relationship with the comfy org team. can ask if they're interested and that would really simplify this!
- thinking of the best way to fold this into the existing example and might fold it into the api cls instead of the ui cls (since that's the use case people care about cold start anyway)

pingren:
@kning thanks for the review!

-  I've actually had some in-person discussions with Comfy Org founders. They're currently prioritizing desktop creators rather than cloud users. Their roadmap for this year is already quite packed. Would you be open to merging this PR first so we can share it with the Modal community, and then we can circle back later with more concrete use cases from Modal?

- This approach could definitely work for the API cls as well. I integrated it into the UI cls mainly for easier testing. I didn't want to overcomplicate the example. Feel free to suggest edits if you think there's a cleaner approach! I am open to make a separate example.

kning:
i think your points make sense! let me try to take an editing pass on it and we can try to merge what we have here.

kning:
@pingren ci is failing so i need to do some work to copy over your forked branch into a branch on this remote and open a new pr to make the tests pass will try to preserve your credit though!

kning:
closing this in favor of https://github.com/modal-labs/modal-examples/pull/1119

PR #1122, opened 2025-04-08
Migrate allow_concurrent_inputs -> modal.concurrent in examples

mwaskom:
Update examples for compatibility with changes to the concurrency configuration in Modal 1.0.

### Type of Change

- [ ] New example

### DIFF
--- a/06_gpu_and_ml/llm-serving/trtllm_latency.py
+++ b/06_gpu_and_ml/llm-serving/trtllm_latency.py
@@ -74,7 +74,9 @@ tensorrt_image = modal.Image.from_registry(
 # including OpenMPI for distributed communication, some core software like `git`,
 # and the `tensorrt_llm` package itself.
 
-tensorrt_image = tensorrt_image.apt_install("openmpi-bin", "libopenmpi-dev", "git", "git-lfs", "wget").pip_install(
+tensorrt_image = tensorrt_image.apt_install(
+    "openmpi-bin", "libopenmpi-dev", "git", "git-lfs", "wget"
+).pip_install(
     "tensorrt-llm==0.18.0",
     "pynvml<12",  # avoid breaking change to pynvml version API
     pre=True,
@@ -101,7 +103,9 @@ tensorrt_image = tensorrt_image.apt_install("openmpi-bin", "libopenmpi-dev", "gi
 # is cached there. And we install `hf-transfer` to get maximum download throughput from
 # the Hugging Face Hub, in the hundreds of megabytes per second.
 
-volume = modal.Volume.from_name("example-trtllm-inference-volume", create_if_missing=True)
+volume = modal.Volume.from_name(
+    "example-trtllm-inference-volume", create_if_missing=True
+)
 VOLUME_PATH = Path("/vol")
 MODELS_PATH = VOLUME_PATH / "models"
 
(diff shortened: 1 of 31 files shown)


mwaskom:
Hm ruff check failures look unrelated?

mwaskom:
Oh well easy fixes I guess.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-17a57c6.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-05d21c7.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9102397.modal.run

06_gpu_and_ml/llm-serving/trtllm_latency.py
```diff
@@ -270,7 +270,7 @@ def get_speculative_config():
 # to process at once before queueing occurs, and the maximum number of prompts
 # to process at once before queueing occurs.
 
-ALLOW_CONCURRENT_INPUTS = 1
+MAX_CONCURRENT_INPUTS = 1
```
mwaskom, line 273: Wasn't totally sure what we were expecting to communicate with this one. Set it to `@modal.concurrent(max_inputs=1)` which is a little weird but not technically illegal.

06_gpu_and_ml/protein-folding/esm3.py
```diff
@@ -223,10 +223,10 @@ def run_esm(sequence: str) -> str:
 
 @app.function(
     image=web_app_image,
-    max_containers=1,  # Gradio requires sticky sessions
-    allow_concurrent_inputs=1000,  # but can handle many async inputs
     volumes={VOLUME_PATH: volume},
+    max_containers=1,  # Gradio requires sticky sessions
 )
+@modal.concurrent(max_inputs=1000)  # but can handle many async inputs
```
mwaskom, line 229: I think this works but the "comment continuation" isn't as clean this way

mwaskom:
@charlesfrye it probably makes sense to demo `target_inputs` in some of these examples. I'm not sure exactly which ones are the best fit (or how best to set the `max`/`target` values. Probably some of the LLM engine ones? I'll let you figure that out if it's ok.

06_gpu_and_ml/llm-serving/trtllm_latency.py
charlesfrye: Ah, the language here could've been more clear.

We are trying to flag for users that the maximum number of concurrent inputs becomes the maximum batch size for the TRT-LLM engine (assuming you don't do multiple samples per request). Will fix.

07_web_endpoints/discord_bot.py
```diff
@@ -53,12 +53,13 @@
 # We convert the response into a Markdown-formatted message.
 
 # We turn our Python function into a Modal Function by attaching the `app.function` decorator.
-# We make the function `async` and set `allow_concurrent_inputs` to a large value because
+# We make the function `async` and add `@modal.concurrent()` with a large `max_inputs` value, because
 # communicating with an external API is a classic case for better performance from asynchronous execution.
```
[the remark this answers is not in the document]
charlesfrye: > "better" performance from async execution and more that you can pile a lot of async tasks onto one thread without any cost.

Hmm, I think of "better performance" here as meaning "reduced latency" -- and running all of these in one thread without async would result in higher latency. It is true that no individual task is executed faster, aside from queueing latency.

charlesfrye:
> @charlesfrye it probably makes sense to demo `target_inputs` in some of these examples. I'm not sure exactly which ones are the best fit (or how best to set the `max`/`target` values. Probably some of the LLM engine ones? I'll let you figure that out if it's ok.

Agreed. Will take a minute to sit down and figure that out -- most of the examples aren't set up for benchmarking/load-testing.

PR #1128, opened 2025-04-16
move from deprecated __init__ to modal.parameter

charlesfrye:
### Type of Change

### DIFF
--- a/06_gpu_and_ml/gpu_packing.py
+++ b/06_gpu_and_ml/gpu_packing.py
@@ -1,3 +1,6 @@
+# ---
+# mypy: ignore-errors
+# ---
 # # Run multiple instances of a model on a single GPU
 #
 # Many models are small enough to fit multiple instances onto a single GPU.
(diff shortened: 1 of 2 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-e891f5f.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-835314d.modal.run

06_gpu_and_ml/gpu_packing.py
```diff
@@ -1,3 +1,6 @@
+# ---
+# mypy: ignore-errors
```
erikbern, line 2: why is this needed?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2d32109.modal.run

06_gpu_and_ml/gpu_packing.py
charlesfrye: Posted in [Slack](https://modal-com.slack.com/archives/C056CGAANRM/p1744770841150439).

Generally dubious of the utility of mypy for the examples -- we can't type-check the logic inside of Functions -- so very quick to turn it off when it gets in the way.

PR #1130, opened 2025-04-16
switch compile parameter to bool

charlesfrye:


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3da209c.modal.run

PR #1132, opened 2025-04-16
minor text fixes

charlesfrye:
We switched from a "comment continuation" to two comments when we added the concurrent decorator in #1122.

I think a parallel structure "Gradio requires ..." and "Gradio can ..." is the right way to do it with two comments.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a2fe6de.modal.run

PR #1133, opened 2025-04-16
update text to clarify relationship between batch and concurrency

charlesfrye:
Also deploys the example, so that the simple chat client in the `__main__` block is always hitting the latest version.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-40b3de7.modal.run

PR #1138, opened 2025-04-23
feat(ollama_deployment)🚀: Add Ollama deployment script for Modal service

ericmjl:
- Create a new Python script to deploy Ollama as a Modal web service.
- Set up a persistent volume for storing model weights.
- Implement a method to pull new models to the service.
- Configure automatic scaling down when not in use.


charlesfrye (approved):
very nice, thanks!

PR #1142, opened 2025-04-26
docs: remove unecessary GPU H100 usage

davidxia:
in simple web endpoint that just returns static text. It's also not relevant for the proxy authentication content.

### Type of Change

- [ ] New example

### DIFF
--- a/07_web_endpoints/basic_web.py
+++ b/07_web_endpoints/basic_web.py
@@ -175,7 +175,7 @@ class WebApp:
 # add the `requires_proxy_auth=True` flag to the `fastapi_endpoint` decorator.
 
 
-@app.function(gpu="h100")
+@app.function()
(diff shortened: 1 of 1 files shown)


luiscape (approved):

charlesfrye:
Thanks for the PR!

The intent of including an expensive resource was to demonstrate that the expensive resource is not actually used when Modal's native proxy authentication fails -- which is not the case if you roll your own authentication.

This is mentioned in the text above, [here](https://github.com/modal-labs/modal-examples/pull/1142/files#diff-1d78eb77c7cfd07e727c2a38bd9a7bd8bd4ee268c85e313e91a92e80a7a5ed40R169-R171).

davidxia:
oh whoops, missed that, thanks!

PR #1152, opened 2025-05-01
adds fastapi to the development requirements

charlesfrye:
A few of the examples `import fastapi` in the global scope, which is important for properly using the `fastapi_endpoint` decorator.

Would prefer to find a different way to solve this (e.g. via an `extra` dependency of `modal`) long-term, but this will prevent all tests from failing while we do so.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-93639ac.modal.run

PR #1157, opened 2025-05-01
WebRTC Demo App using aiortc and YOLO

shababo:
### Summary

This PR introduces a new example which demonstrates how to utilize WebRTC with Modal to perform real-time object detection on client's webcam stream with YOLO.

### App Design

There are three main components to the app:

**Client/Stream Provider Peer:** Source of media stream and negotiation initiator via a websocket (browser frontend or `WebRTCVideoProcessorTester` container on Modal).

**Modal WebRTC Server:** Provides `fastapi.WebSocket` endpoint for client, spawns the stream processing peer (Modal function) with a `modal.Queue`, and relays negotiation messages between the `WebSocket` and `Queue`

**Stream Processing Peer:** Processes an incoming media stream and sends the stream back to the provider

With this design, the lifecycle of the Stream Processing Peer can be tied to a single Modal FunctionCall which allows developers to take advantage of Modal auto-scaling as a function of Streams/Peer Connections. Additionally, this decouples the P2P connection lifetime from the lifetime of web-based connections between the client and server. In fact, the Websocket is closed once the WebRTC P2P connection is established.

#### Implementation Details/NBs
- There are two base classes, `ModalWebRTCPeer` and `ModalWebRTCServer` which encapsulate a lot of the WebRTC/`aiortc` complexity/boiler plate and provide the user with overridable methods for initializing/shutting down the peer, setting up the streams, etc.
- It is possible to implement this without the intermediary server which reduces negotiation time and eliminates delays due to spinning up new containers. However the Server enables more flexible architectures/auto-scaling. Additionally, when using a GPU instance and loading models/etc, the negotiation time difference becomes less relevant.
- The browser frontend allows the user to select STUN or TURN servers. For the former we use Google's servers. For the latter, @shababo signed up for a free TURN server service (up to 500 GB). TURN is necessary at the SF office and in many other network scenarios. Mercifully, Modal containers seem happy to use STUN.
- We use [Trickle Ice](https://datatracker.ietf.org/doc/html/rfc8838) to send candidates from the client peer to the modal peer. Without it, the negotiation can take on the order of minutes. Again, mercifully, we don't need to use this to send Modal peer + STUN candidates back to the client.
- YOLO is using the `onnxruntime` with `tensorrt`. To enable faster spin up, we are caching model weights and the ONNX inference session graph to a volume. We also need to deal with some `locale` stuff when we build the container image which improves inference times from 4-5 msec to 2-3 msec per frame.

### Testing
The app includes a `modal.Cls` that implements `ModalWebRTCPeer`. This class plays a video over WebRTC to the Modal Peer and records the returned processed video to a file. It then ensures the new file is no more than 5 frames shorter than the original source video file.

#### Potential Improvements
- use `model.parameters` with deployed stream processing functions to dynamically choose how to process the video (e.g. obj detection, pose, etc.) when `spawn`ing
- we could set a modal peer container to buffer when the client starts their webcam to reduce time between hitting the streaming button and start of the stream; at the very least we could add something in the GUI to let user's know the function/container or spinning up

#### Misc
- should we use a logger instead of print statements?
- test working with audio

### Type of Change

- [x] New example

## Checklist


07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
```diff
@@ -0,0 +1,420 @@
...
+
+# instantiate our app
+app = modal.App(APP_NAME)
+
+
+# our modal peer, a subclass of ModalWebRTCPeer
+# this class flips and incoming video stream
+# and then streams the flipped video back to the provider
+@app.cls(
+    image=video_processing_gpu_image,
```
aksh-at, line 85: Worth mentioning region pinning as an option here, so you can control the latency between the client and the GPU container.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
```diff
@@ -0,0 +1,420 @@
...
+
+        onnxruntime.preload_dlls()
+        self.yolo_model = YOLOv10(CACHE_PATH)
+
+    async def setup_streams(self, peer_id: str):
+        import numpy as np
+        from aiortc import MediaStreamTrack
+        from aiortc.contrib.media import VideoFrame
+
+        class YOLOTrack(MediaStreamTrack):
```
aksh-at, line 110: Does this have to be nested within `setup_streams`?

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
```diff
@@ -0,0 +1,420 @@
...
+            return HTMLResponse(content=html)
+
+        return self.web_app
+
+
+# create an output volume to store the transmitted videos
+
+
+@app.cls(image=tester_image, volumes={CACHE_PATH: CACHE_VOLUME})
+class WebRTCVideoProcessorTester(ModalWebRTCPeer):
```
aksh-at, line 267: Should this be in a separate file?

07_web_endpoints/aiortc/modal_webrtc/modal_webrtc.py
```diff
@@ -0,0 +1,485 @@
...
+    Class that handles WebRTC signaling
+    between WebRTC clients and a modal app
+    that implements the WebRTCPeer class
+    """
+
+    # not using this, but could use it to spawn
+    # user defined stream processing functions
+    # however this only works if the modal app is
+    # already deployed
+    modal_peer_app_name: str = modal.parameter()
```
aksh-at, line 27: I think in general it's nice to have the peer function be in the same app. Probably worth removing this to not confuse people.

07_web_endpoints/aiortc/modal_webrtc/modal_webrtc.py
```diff
@@ -0,0 +1,485 @@
...
+@dataclass
+class IceCandidate:
+    peer_id: str
+    candidate_sdp: str
+    sdpMid: str
+    sdpMLineIndex: int
+    usernameFragment: str
+
+
+class ModalWebRTCServer:
```
aksh-at, line 16: nit: `ModalWebRtcServer`

07_web_endpoints/aiortc/modal_webrtc/modal_webrtc.py
```diff
@@ -0,0 +1,485 @@
...
+                            )
+                            return
+
+            await asyncio.gather(
+                relay_client_messages(client_websocket, q),
+                relay_modal_peer_messages(client_websocket, q),
+            )
+
+
+class ModalWebRTCPeer:
```
aksh-at, line 124: nit: `ModalWebRtcPeer`

aksh-at (commented):
Nice! Works well. The `ModalWebRtcServer` and `Peer` concepts seem pretty useful, perhaps we could bring them into the client one day. We'll have to think about how to make the interface really solid though. There's some things that seem a bit leaky right now.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
shababo: Moved into `.initialize`

07_web_endpoints/aiortc/modal_webrtc/modal_webrtc.py
```diff
@@ -0,0 +1,485 @@
...
+    # we use this classvar instead since we have access to it
+    modal_peer_cls: ClassVar = None
+
+    @modal.enter()
+    def initialize(self):
+        import uuid
+
+        from fastapi import FastAPI, WebSocket
+
+        # no longer used but seems worth keeping around
```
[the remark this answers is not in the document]
shababo: See previous comment for relevant thread.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
shababo: Defer to you and @charlesfrye on this... initially there was an instruction to keep examples to a single .py file for testing reasons.

07_web_endpoints/aiortc/modal_webrtc/modal_webrtc.py
shababo: Taking this out.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
aksh-at: I actually meant it can be a class in global scope.

aksh-at:
@shababo also just noticed we're checking in `cliff_jumping.mp4`. Can we skip that? The examples repo is pretty bloated with old assets as is (we need to clean some of that up).

shababo:
> @shababo also just noticed we're checking in `cliff_jumping.mp4`. Can we skip that? The examples repo is pretty bloated with old assets as is (we need to clean some of that up).

Definitely. I had been hesitant to commit a binary file. Do we have a way to serve files or maybe to create publicly accessible (read only) modal.Volumes or something like that?

charlesfrye:
I told Ben he could include the video binary.

@shababo current practice is to put in Modal's public assets in S3 and access via the `modal-cdn.com` URL. I can do that in this case if you make the logic work off of a URL.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
aksh-at: It's just the types that are in global scope then rihgt? You can put them in quotes to avoid them being needed locally I believe.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
charlesfrye: We can put it at the end of the file with clear language indicating that it's an extra, for testing purposes.

This is a pattern I quite like with Modal -- local entrypoint for a deployed/served web app just does tests, so you can do quick iterative development via `modal run`.

07_web_endpoints/aiortc/modal_webrtc/webrtc_yolo.py
charlesfrye: The class is inheriting from `MediaStreamTrack` in `aiortc`.

aksh-at: Ah right of course. Yeah it's fine in that case then.

charlesfrye:
https://modal-cdn.com/cliff_jumping.mp4 @shababo

shababo:
> https://modal-cdn.com/cliff_jumping.mp4

success.

shababo:
Closing this PR since we're back in "dev" mode in terms of writing the copy. Will open a new one when it's fully ready for review.

PR #1167, opened 2025-05-09
speed up ltx-video and reorganize text

charlesfrye:
Explicitly moved the pipeline to the GPU, dropped the CPU offload, and set the dtype to bf16. This cuts inference time down substantially -- for a twenty second 480p video, can complete 10 steps in 2s or 100 steps in 20s.

Added the `--twice` flag/behavior we use in a few examples where we want to demonstrate the size of cold start latencies.

Reorganized the text a bit. Added more links to other Modal docs.

Removed the `download_model` function that didn't appear to be used. It's a good pattern, but we should put it in the model weights storage guide (also now linked in the example).

### Type of Change

### DIFF
--- a/06_gpu_and_ml/text-to-video/ltx.py
+++ b/06_gpu_and_ml/text-to-video/ltx.py
@@ -1,14 +1,10 @@
-# ---
-# cmd: ["modal", "run", "--detach", "06_gpu_and_ml/text-to-video/ltx.py", "--num-inference-steps", "64"]
-# ---
+# # Generate videos from prompts with Lightricks LTX-Video
 
-# # Text-to-video generation with Lightricks LTX-Video
-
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3b1959a.modal.run

charlesfrye:
We could go even faster, including during download and cold boot, if we got their [fp8-quantized model ](https://huggingface.co/Lightricks/LTX-Video/blob/main/ltxv-13b-0.9.7-dev-fp8.safetensors) to run. Leaving that one to you, @zhang-lucy!

zhang-lucy:
Sweeeet thanks for the improvements here @charlesfrye ! Will take a stab at the super speedy model 👀

PR #1174, opened 2025-05-13
Parakeet real time streaming APE-6

zhang-lucy:
https://linear.app/modal-labs/issue/APE-6/parakeet-websocket-example

### Type of Change

- [x] New example

## Checklist

  - [] Example is tested by executing with `modal run` or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "deploy"]`)
- [ ] Example does _not_ require third-party dependencies to be installed locally, except `fastapi` - [followup](https://linear.app/modal-labs/issue/APE-8/parakeet-demo-browser-webapp)

## Outside contributors


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-009c245.modal.run

zhang-lucy:
@charlesfrye - still working on this one, but would love your thoughts on how to best structure this example. It's simple enough that I'd like it to be a standalone example, but it does have 2 parts - client/server, where the client is not deployed using Modal

[more context here](https://modal-com.slack.com/archives/C08AK4B7S81/p1747161040063979?thread_ts=1747146367.497909&cid=C08AK4B7S81)

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d24cca0.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,238 @@
...
+                while True:
+                    chunk = await ws.receive_bytes()
+                    buffer.extend(chunk)
+
+                    if len(buffer) > BUFFER_SIZE:
+                        audio_np = np.frombuffer(bytes(buffer), dtype=np.int16)
+                        buffer.clear()
+
+                        try:
+                            text = self.transcribe.remote(audio_np)
```
luiscape, line 129: Calling `text = self.transcribe.remote(audio_np)` from this websocket will call this in a _different_  container. This can be fine if expect a container to work as a dispatcher and one to do the GPU processing. That doesn't seem to be the case here, though. Use `text = self.transcribe.local(audio_np)` instead.

zhang-lucy: got it, ty!

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2f1a071.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-70bd5d1.modal.run

06_gpu_and_ml/audio-to-text/parakeet_client.py
```diff
@@ -0,0 +1,96 @@
+# ---
+# lambda-test: false
```
zhang-lucy, line 2: not sure why the tests are still running

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+            "HF_HUB_ENABLE_HF_TRANSFER": "1",
+            "HF_HOME": "/cache",  # cache directory for Hugging Face models
+            "DEBIAN_FRONTEND": "noninteractive",
+            "CXX": "g++",
+            "CC": "g++",
+        }
+    )
+    .apt_install("ffmpeg")
+    .run_commands(
+        "uv pip install --system hf_transfer huggingface_hub[hf-xet] nemo_toolkit[asr] cuda-python>=12.3",
```
zhang-lucy, line 67: will pin these deps

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d9ccb95.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a791c7c.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f3d97ec.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b2955a2.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-196ed1d.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4e2f75f.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ecc3890.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+            "HF_HUB_ENABLE_HF_TRANSFER": "1",
+            "HF_HOME": "/cache",  # cache directory for Hugging Face models
+            "DEBIAN_FRONTEND": "noninteractive",
+            "CXX": "g++",
+            "CC": "g++",
+        }
+    )
+    .apt_install("ffmpeg")
+    .run_commands(
+        "uv pip install --system hf_transfer==0.1.9 huggingface_hub[hf-xet]==0.31.2 nemo_toolkit[asr]==2.3.0 cuda-python==12.9.0",
```
charlesfrye, line 82: Curious -- why are we using uv pip install here? And in multiple steps?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a95e3fd.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+# For inference, we recommend using the official NVIDIA CUDA Docker images from Docker Hub.
+# You'll need to add Python 3 and pip with the `add_python` option because the image
+# doesn't have these by default.
+
+# Additionally, we install `ffmpeg` for handling audio data and `fastapi` to create a web
+# server for our websocket. `uv` gives us even more speed on installation!
+
+image = (
+    modal.Image.from_registry(
+        "nvidia/cuda:12.8.0-cudnn-devel-ubuntu22.04", add_python="3.12"
```
charlesfrye, line 68: The version here is 12.8 and the cuda-python version differs -- previously 12.3, now 12.9.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a4aeb64.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+# We'll use  python's `websockets` library to create a websocket client
+# that sends audio data to the server and receives transcriptions in real-time.
+
+
+client_image = (
+    modal.Image.debian_slim(python_version="3.10")
+    .apt_install("portaudio19-dev")
+    .pip_install("websockets==15.0.1", "sounddevice==0.5.1", "numpy==2.2.5")
+)
+client_app = modal.App("parakeet-client", image=client_image)
```
charlesfrye, line 173: Why two apps?

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+        async with websockets.connect(
+            ws_url, open_timeout=240, ping_interval=None
+        ) as websocket:
+            send_task = asyncio.create_task(send_audio(websocket, audio_queue))
+            receive_task = asyncio.create_task(receive_transcriptions(websocket))
+            await asyncio.gather(send_task, receive_task)
+
+    is_dev = True  # set to False if running modal deploy
+
+    url = f"wss://{modal_profile}--{app_name}-{class_name}-web{'-dev' if is_dev else ''}.modal.run"
```
charlesfrye, line 230: I think you can construct this with `.get_web_url`

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,246 @@
...
+CHUNK_DURATION = 1.0
+CHUNK_SIZE = int(SAMPLE_RATE * CHUNK_DURATION)
+DTYPE = "int16"  # must be string for deferred eval
+
+
+@client_app.local_entrypoint()
+def main(modal_profile: str):
+    import asyncio
+
+    import sounddevice as sd
```
charlesfrye, line 188: Flagging that this still requires sounddevice as a local dependency

charlesfrye (changes requested):

06_gpu_and_ml/audio-to-text/parakeet.py
zhang-lucy: not sure how to do this with one app - we need to modal serve the server and run the client locally - taking suggestions?

06_gpu_and_ml/audio-to-text/parakeet.py
luiscape: > why are we using uv pip install here?

Because it's faster than pip. Maybe I'm not understanding the question. 

> And in multiple steps?

The additional numpy step is there to downgrade numpy. The current numpy version installed with NeMo is incompatible with this model.

charlesfrye: > why are we using uv pip install here?

I don't think we usually use `uv` in examples, even though it is faster, because we don't have an interface for it, so it feels a bit janky: `run_commands([...])` instead of `pip_install("", ...)`.

> The additional numpy step is there to downgrade numpy. The current numpy version installed with NeMo is incompatible with this model.

Why can't we do `numpy<2` within the same step? And I don't think we need a third step to install `fastapi`.

Asking because `pip` doesn't take into account dependency resolution across invocations, and I don't think `uv pip` does either.

charlesfrye: Also I think some of the speedup may be illusory because deferred to a runtime expense -- by default, `uv` does not compile bytecode for packages. cf [Image guide](https://modal.com/docs/guide/images#run-shell-commands-with-run_commands) in the docs

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4541fa7.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
zhang-lucy: removed uv

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-21bef2d.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,282 @@
...
+    import requests
+
+    CHUNK_SIZE = 64000
+    is_dev = True
+    url = f"wss://{modal_profile}--{app_name}-{class_name}-web{'-dev' if is_dev else ''}.modal.run"
+    ws_url = f"{url}{WS_ENDPOINT}"
+
+    def convert_to_mono_16khz(audio_bytes: bytes) -> bytes:
+        import io
+        import wave
```
charlesfrye, line 197: nice! TIL

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,282 @@
...
+WS_ENDPOINT = "/ws"
+
+AUDIO_URL = "https://github.com/voxserv/audio_quality_testing_samples/raw/refs/heads/master/mono_44100/156550__acclivity__a-dream-within-a-dream.wav"
+
+
+@client_app.local_entrypoint()
+def main(modal_profile: str, audio_url: str = AUDIO_URL):
+    import asyncio
+
+    import requests
```
charlesfrye, line 188: `requests` is not a dependency of `modal`, so we can't use it. Will provide suggested alternatives.

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,282 @@
...
+
+        async with websockets.connect(
+            ws_url, ping_interval=None, open_timeout=240
+        ) as ws:
+            send_task = asyncio.create_task(send_audio(ws, audio_bytes))
+            receive_task = asyncio.create_task(receive_transcriptions(ws))
+            await asyncio.gather(send_task, receive_task)
+
+    print("🌐 Downloading audio file...")
+    response = requests.get(audio_url)
```
charlesfrye, line 261: We can use [`urllib.request.urlopen`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen) here, and it's actually pretty clean! See [this sample](https://github.com/modal-labs/modal-examples/blob/063315bf657b86d0bb3785bf9017931c6149dfaa/06_gpu_and_ml/llm-serving/chat_with_pdf_vision.py#L359).

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,282 @@
+# ---
+# cmd: ["modal", "serve", "06_gpu_and_ml/audio-to-text/parakeet.py::server_app"]
```
charlesfrye, line 2: If you collapse this into one app, as described below, this ends up becoming just `modal run`, which is the default `cmd`, and so you can elide this field -- and the entire frontmatter!

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -0,0 +1,282 @@
...
+        for chunk in chunk_audio(audio_bytes, CHUNK_SIZE):
+            await ws.send(chunk)
+            await asyncio.sleep(0.01)  # simulate real-time pacing
+
+    async def receive_transcriptions(ws):
+        async for message in ws:
+            print("📝 Transcription:", message)
+
+    async def run(ws_url, audio_bytes):
+        import websockets
```
charlesfrye, line 251: Unfortunately I don't think that there's a stdlib implementation of `websockets`, so I'm not sure how we're going to implement this.

charlesfrye (changes requested):
Our usual requirement for the examples is that they have no local dependencies besides `modal` -- and `fastapi` if they use `fastapi_endpoint` features.

This is necessary to support examples QA -- we need to control the maintenance effort and the complexity of our examples monitoring.

Alternatively, we can just set `lambda-test: false` in the frontmatter and come up with a way to test it later, as part of the planned next step of adding a browser frontend.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3b8d1a0.modal.run

charlesfrye (approved):

charlesfrye:
good to merge!

PR #1187, opened 2025-05-21
Parakeet example frontend, remove local deps, and other fixes

shababo:
Made the following changes to the Parakeet example:
- Added a frontend that was very much "vibe coded." Hopefully nothing too crazy in there.
- Made example contained in a single app.
- Added concurrency to allow multiple clients and be more robust to stop/restarting.
- `local_entrypoint` no longer uses non standard/non Modal dependencies. Using a `Queue` to send audio and transcriptions back and forth.
- Properly end calls to Modal app when client disconnects or wav file finishes playing
- Detect silence and use to chunk audio into phrases/sentences for better transcription. Parakeet wants to output properly capitalized and punctuated English text - so if you give it partial chunks or chunks with no audio it will produce "hallucinations"/incorrect output.

## Type of Change

### ☑️ Monitoring
  - [x] Example is configured for testing in the synthetic monitoring system
    - [x] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or in the scope of a `local_entrypoint`)

### ☑️ Content
  - [ ] Example media assets are retrieved from `modal-cdn.com`

### ☑️ Build Stability

## Outside Contributors


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a9566b2.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -10,17 +10,15 @@
...
-# modal serve 06_gpu_and_ml/audio-to-text/parakeet.py::server_app
+# modal serve -m 06_gpu_and_ml.audio-to-text.parakeet.py
 # ```
-
-# 2. In a separate terminal, run the client:
+# B. Stream a .wav file from URL (default is "Dream Within a Dream" by Edgar Allan Poe):
 # ```bash
-# pip install requests websockets numpy
-# modal run 06_gpu_and_ml/audio-to-text/parakeet.py::client_app --modal-profile=$(modal profile current)
+# modal run -m 06_gpu_and_ml.audio-to-text.parakeet --audio-url=<URL_TO_WAV_FILE>
```
zhang-lucy, line 21: can we add the URL here so it's copy-pasteable?

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -31,30 +29,32 @@
 
 # 🌐 Downloading audio file...
 # 🎧 Downloaded 6331478 bytes
-# 🔗 Streaming data to WebSocket:
-# wss://modal-labs--parakeet-websocket-parakeet-web-dev.modal.run/ws
+# 🔗 Streaming data to WebSocket: wss://modal-labs-shababo-dev--parakeet-websocket-parakeet-web-dev.modal.run/ws
```
zhang-lucy, line 32: nit: we should deploy this on main and point the URL to that, on the offchance a user actually tries to hit this

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -31,30 +29,32 @@
...
+SILENCE_THRESHOLD = -45
+SILENCE_MIN_LENGTH_MSEC = 1000
+END_OF_STREAM = b"END_OF_STREAM"
 # ## Volume for caching model weights
 # We use a [Modal Volume](https://modal.com/docs/guide/volumes) to cache the model weights.
 # This allows us to avoid downloading the model weights every time we start a new instance.
 
 
-model_cache = modal.Volume.from_name("parakeet-model-cache", create_if_missing=True)
+model_cache = modal.Volume.from_name("parakeet-model-cache-2", create_if_missing=True)
```
zhang-lucy, line 57: nit
```suggestion
model_cache = modal.Volume.from_name("parakeet-model-cache", create_if_missing=True)
```

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
+# - The `transcribe` method takes bytes of audio data, and returns the transcribed text.
 # - The `web` method creates a FastAPI app using [`modal.asgi_app`](https://modal.com/docs/reference/modal.asgi_app#modalasgi_app) that serves a
-# [websocket](https://modal.com/docs/guide/webhooks#websockets) endpoint for real-time audio transcription.
-
-# Additionally, since this app is running in its own container, we make a `.remote` call to invoke
-# the `transcribe` method. This allows us to use the GPU for inference.
+# [websocket](https://modal.com/docs/guide/webhooks#websockets) endpoint for real-time audio transcription and a browserfrontend for transcribing audio from your microphone.
+# - Parakeet is designed to output properly punctuated/capitalized English text. Therefore it does not perform well on chunks of audio that are not well-formed sentences.
+# And it also will output ("hallucinate") utterances like "Yeah" or "Mm-hmm" when it runs on silent audio. Therefore we group the incoming audio in the server based on silenece detection
+# using `pydub`.
```
zhang-lucy, line 111:
```suggestion
# Parakeet tries really hard to transcribe everything to English! Hence it tends to output utterances like "Yeah" or "Mm-hmm" when it runs on silent audio. We can pre-process the incoming audio in the server by using `pydub`'s silence detection, ensuring that we only pass audio with text to our model.
```

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
     def load(self):
+        import nemo.collections.asr as nemo_asr
+
         self.model = nemo_asr.models.ASRModel.from_pretrained(
             model_name="nvidia/parakeet-tdt-0.6b-v2"
         )
 
-    @modal.method()
-    def transcribe(self, audio_bytes: bytes) -> str:
+    # @modal.method()
```
zhang-lucy, line 127: nit
```suggestion
```

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
 
-client_image = modal.Image.debian_slim(python_version="3.12")
-client_app = modal.App("parakeet-client", image=client_image)
 
+# ## Client
+# Next, let's test the model with a `local_entrypoint` that streams audio data to the server, and prints
+# out the transcriptions in real-time to our terminal. We can also run this using Modal!
 
-WS_ENDPOINT = "/ws"
+# We'll use modal's `Queue` to pass audio data and transcriptions between your local machine and the server.
```
zhang-lucy, line 254:
```suggestion
# We'll use a [`Modal Queue`](https://modal.com/docs/reference/modal.Queue) to pass audio data and transcriptions between your local machine and the server.
```

06_gpu_and_ml/audio-to-text/parakeet.py
zhang-lucy: or have it default to that URL, and add a comment that they can pass in a diff one

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
+        last_window = silent_windows[-1]
+        # if the entire audio segment is silent, reset the audio segment
+        if last_window[0] == 0 and last_window[1] == len(audio_segment):
+            audio_segment = AudioSegment.empty()
+            return audio_segment, None
+        # get the segment to transcribe: beginning until last pause
+        segment_to_transcribe = audio_segment[: last_window[1]]
+        # remove the segment to transcribe from the audio segment
+        audio_segment = audio_segment[last_window[1] :]
+        try:
```
zhang-lucy, line 242: might be missing something, but does this handle <audio>, <silence>, <audio>, <silence>? aka multiple silences

zhang-lucy (approved):
🔥

06_gpu_and_ml/audio-to-text/parakeet.py
charlesfrye: Agreed! And you shouldn't need the `-m` here. I forget the exact mechanics, but it should only be needed if you're importing other Python code as modules.

06_gpu_and_ml/audio-to-text/parakeet.py
charlesfrye: Deploying to main for URLs we include in the text is an interesting idea! I think it'd be a good practice for`https` URLs.

But this is a `wss` URL, which I don't think you can hit via a browser? Or at least you shouldn't try.

Also an FYI, but the environment we use for examples is `examples`, not `main`. I don't think that's documented anywhere, sorry! The `main` environment has a bunch of Apps and Secrets used by the core platform, and those have a higher SLO for reliability than examples do.

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
 
 # - The `load` method loads the model at start, instead of during inference, using [`modal.enter()`](https://modal.com/docs/reference/modal.enter#modalenter).
-# - The `transcribe` method takes a numpy array of audio data, and returns the transcribed text.
+# - The `transcribe` method takes bytes of audio data, and returns the transcribed text.
 # - The `web` method creates a FastAPI app using [`modal.asgi_app`](https://modal.com/docs/reference/modal.asgi_app#modalasgi_app) that serves a
-# [websocket](https://modal.com/docs/guide/webhooks#websockets) endpoint for real-time audio transcription.
-
-# Additionally, since this app is running in its own container, we make a `.remote` call to invoke
-# the `transcribe` method. This allows us to use the GPU for inference.
+# [websocket](https://modal.com/docs/guide/webhooks#websockets) endpoint for real-time audio transcription and a browserfrontend for transcribing audio from your microphone.
```
charlesfrye, line 108:
```suggestion
# [websocket](https://modal.com/docs/guide/webhooks#websockets) endpoint for real-time audio transcription and a browser frontend for transcribing audio from your microphone.
```

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
 AUDIO_URL = "https://github.com/voxserv/audio_quality_testing_samples/raw/refs/heads/master/mono_44100/156550__acclivity__a-dream-within-a-dream.wav"
+TARGET_SAMPLE_RATE = 16000
+CHUNK_SIZE = 16000  # send one second of audio at a time
 
 
-@client_app.local_entrypoint()
-def main(modal_profile: str, audio_url: str = AUDIO_URL):
+@app.local_entrypoint()
+def main(audio_url: str = AUDIO_URL):
+    import array
```
charlesfrye, line 263: nice

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +87,279 @@
...
+# We'll use modal's `Queue` to pass audio data and transcriptions between your local machine and the server.
 
 AUDIO_URL = "https://github.com/voxserv/audio_quality_testing_samples/raw/refs/heads/master/mono_44100/156550__acclivity__a-dream-within-a-dream.wav"
+TARGET_SAMPLE_RATE = 16000
+CHUNK_SIZE = 16000  # send one second of audio at a time
 
 
-@client_app.local_entrypoint()
-def main(modal_profile: str, audio_url: str = AUDIO_URL):
+@app.local_entrypoint()
```
charlesfrye, line 261: This function is pretty long -- almost 100 lines. We should break it up.

Luckily that doesn't look too hard. It's mostly composed of function definitions and I think all of the functions defined here can instead be defined in the global scope, because they are valid locally and remotely.

For `send_audio`, `receive_transcription`, and `run`: It's worth saying a few words before their definition about Queues (but deferring details to the relevant Guide page, i.e. the URL @zhang-lucy added) and in particular why we're using them here. That's a Modal feature and so something we are on the hook for documenting.

For `preprocess_audio` and `chunk_audio`: these strike me as mostly boilerplate -- unless I'm missing something. Toss those two in a section called "Addenda" at the end (cf other examples).

06_gpu_and_ml/audio-to-text/parakeet.py
charlesfrye, line 1: Looks like this now works with `modal run` with only dependencies available in testing, so we should be able to delete the `lambda-test` key -- and so the entire YAML frontmatter.

shababo: ope. this was an uncommited change i had made :( pushed now though! :)

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: Modal keeps yelling at me that in the future I will need `-m`...
```bash
╭─ Modal Deprecation Warning (2025-02-06) ─────────────────────────────────────────────────────────╮
│ Using Python module paths will require using the -m flag in a future version of Modal.           │
│ Use `modal serve -m parakeet` instead.                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

shababo: It already does default to that URL. Added it for clarity though.

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: Ope... this lines don't get printed anymore anyway because we use the Queue and not Websockets for the local entrypoint. We don't have to worry about the URl stuff anymore anyway because it's all in one app so we can use `get_web_url()` if we were still using websockets here.

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: Good catch! Leftover from a test!

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: Yes. It's basically goes to the last detected silence in the segment and transcribes everything up to there. Then it keeps what's left (some amount of talking that doesn't stop) and will concat the next incoming segment to that one for the next loop.

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: not a nit! this should definitely be deleted!

06_gpu_and_ml/audio-to-text/parakeet.py
charlesfrye: > Added it for clarity though.

Either works -- copy-pastable or clearly extensible. I think we do different things in different places.

> Modal keeps yelling at me that in the future I will need -m

Hmm. Are you including `.py` at the end of the filename? Are you using `/` separators in the path? That should be enough to run as a script instead of a module.

shababo:
```bash
(modal-examples) shababo@Bens-MacBook-Air audio-to-text % modal run parakeet
╭─ Modal Deprecation Warning (2025-02-06) ─────────────────────────────────────────────────────────╮
│ Using Python module paths will require using the -m flag in a future version of Modal.           │
│ Use `modal run -m parakeet` instead.                                                             │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
```

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4f969be.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
charlesfrye: Sorry I should've clarified what the "right" answers are there. ` modal run parakeet.py` or `modal run path/to/parakeet.py` should work with no warnings.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-2850e1f.modal.run

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: Ohhh... I suppose I forgot that you add the file extension when not using `-m` after only using the flag for the WebRTC example. I'll update the copy to match.

06_gpu_and_ml/audio-to-text/parakeet.py
```diff
@@ -87,194 +86,312 @@
...
+# `send_audio` transmits chunks of audio data and then pauses to approximate streaming
+# speech at a natural rate. That said, we set it to faster
+# than real-time to compensate for network latency. Plus, we're not
+# trying to wait forever for this to finish.
 
 
-@client_app.local_entrypoint()
-def main(modal_profile: str, audio_url: str = AUDIO_URL):
+async def send_audio(q, audio_bytes):
     import asyncio
```
charlesfrye, line 289: non-blocking: `asyncio` is `stdlib`, so you can import it in the global scope at the top. especially nice because you end up importing it in multiple small functions here!

charlesfrye (approved):

charlesfrye:
LGTM! Nice work.

06_gpu_and_ml/audio-to-text/parakeet.py
shababo: You're referring to just `asyncio` and not the other stdlib imports, yeah? I like them being local for pedagogical reasons but `asyncio` would be reasonable to move.

charlesfrye: yeah, focusing on it in particular because it's imported repeatedly and in smaller function bodies. I agree that moving the import out of the block at the top of the file can be nice for the prose/flow!

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a57537a.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-11c4b54.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d55aae9.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b4a944e.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9411c84.modal.run

PR #1168, opened 2025-05-09
removes fetch_stock_prices

charlesfrye:
The fetch_stock_prices example breaks fairly regularly due to web scraping issues and changes to the Yahoo! Finance API. It isn't referenced in many places and we have other analytics and scraping examples. Dependent on removal of references in the docs site ([PR](https://github.com/modal-labs/modal/pull/22429)).

PR #1194, opened 2025-05-22
fix broken link

zhang-lucy:
## Type of Change

Would be so cool if we could add some automation for this, but alas probably overkill

### DIFF
--- a/01_getting_started/hello_world.py
+++ b/01_getting_started/hello_world.py
@@ -118,7 +118,7 @@ def main():
 # or [fine-tuning](https://modal.com/docs/examples/slack-finetune)
 # * Manipulating [audio](https://modal.com/docs/examples/musicgen)
 # or [images](https://modal.com/docs/examples/diffusers_lora_finetune)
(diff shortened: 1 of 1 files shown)


charlesfrye:
thanks! deprecated that example in #1168 but missed the link here. i think i grepped frontend, forgetting that you also need to grep examples

we do detect all broken links in other pages on the site! but this repo is intended to be standalone (you can click links from inside VS Code, for instance!) and so we don't benefit from the same behavior (iirc from the server-side rendering)

a simple crawler that checks all modal.com-prefixed URLs in the repo for 404s would be a fun example to work on!

charlesfrye (approved):

zhang-lucy:
> thanks! deprecated that example in #1168 but missed the link here. i think i grepped frontend, forgetting that you also need to grep examples
> 
> we do detect all broken links in other pages on the site! but this repo is intended to be standalone (you can click links from inside VS Code, for instance!) and so we don't benefit from the same behavior (iirc from the server-side rendering)
> 
> a simple crawler that checks all modal.com-prefixed URLs in the repo for 404s would be a fun example to work on!

sickk always awesome to learn about new automations that y'all have set up!

PR #1197, opened 2025-05-23
resolve or document monitoring opt-outs

charlesfrye:
Removes or documents all of the monitoring opt-outs. Those I fixed are very old opt-outs from when we couldn't do as much.

I used the following slugs to document reasons in YAML comments:
- `auxiliary-file`: this is an auxiliary Python file that's part of an example but not a monitored unit in its own right, like a script to be run from a Modal Function. This is the only good reason to skip monitoring, but it should still be used sparingly. A self-contained, self-documented application is a beautiful thing!
- `long-running`: this example runs for too long. These are all in `12_datasets`. We should rethink these examples in light of notebooks, cc @thundergolfer.
-  `deprecated`: this example should be removed. They're waiting on the deployment of a PR to `modal` [here](https://github.com/modal-labs/modal/pull/22917)
- `missing-secret`: this example uses a Secret that was deleted. These examples should be resuscitated or removed.

## Type of Change

### DIFF
--- a/05_scheduling/hackernews_alerts.py
+++ b/05_scheduling/hackernews_alerts.py
@@ -1,5 +1,5 @@
 # ---
-# lambda-test: false
+# lambda-test: false  # missing-secret
 # ---
 
 # # Run cron jobs in the cloud to search Hacker News
--- a/06_gpu_and_ml/llm-structured/jsonformer_generate.py
+++ b/06_gpu_and_ml/llm-structured/jsonformer_generate.py
@@ -1,5 +1,5 @@
 # ---
-# lambda-test: false
+# lambda-test: false  # deprecated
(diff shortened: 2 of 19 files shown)


charlesfrye:
@prbot approve

charlesfrye:
@prbot automerge

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ca862ac.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9934ae5.modal.run

wshainin (commented):
LGTM

PR #1214, opened 2025-06-10
Fix broken 'create secret' link

paulgb:
The Hacker News example has a link that leads to a 404. Since the create secret link seems to depend on knowing the account ID, I've replaced it with instructions.

image

## Type of Change


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-0751efa.modal.run

PR #1217, opened 2025-06-12
update vllm example

charlesfrye:
### What if the vLLM example was fast?

This PR makes a number of changes to make the vLLM inference example run and feel faster.

It reduces the cold start time from over four minutes to less than one while generating more tokens and getting them to the client faster.

### How does this PR make the vLLM example faster?

- It switches to eager mode by default. vLLM has generally moved in the direction of trading compile time for run time, which is sensible in a large, stable production environment but isn't great for a demo of a serverless platform. Their docs incorrectly say that the default compilation level is `-O0`, but from my experiments it's actually `-O3`. I gave up on tuning the flags more tightly than just `eager`/`no-eager` because I couldn't get them to be respected, and it turns out `--enforce-eager` turns off pretty much everything, which is basically what we want. This PR also documents the choice and makes it more obviously configurable. **This is the biggest change**, cutting ~145s of latency on cold boot.
- It updates vLLM to 0.9, which has a number of perf improvements, including running a newer Torch version.
- It moves to Blackwell, since we have Blackwells and Blackwell is supported in vLLM/Torch and in flashinfer (requires CUDA 12.8). We can even install flashinfer with CUDA 12.8 from a wheel now! This has only a small impact on speed, because it doesn't impact cold start and inference is firmly in the host overhead regime.
- It adds streaming responses, so the first token shows up faster. Latency is perceived, not just measured!
- It adjusts the testing so that more work is done, which paradoxically makes it _feel_ faster -- the boot is amortized over more work. It adjusts the prompt so that more tokens are produced and sets a new default behavior of sending two user messages (with different system prompts).

### Why isn't this even faster?

Well, what are we spending our ~45s of boot time on? Let's look at the sample logs [here](https://modal.com/apps/modal-labs/examples/ap-S50cxoTULd1AkgIZMdl7Gy?start=1749693451.135&end=1749697051.135&live=true&fcId=fc-01JXGZWJH0HRJGF78VMT2SM278&activeTab=functions&functionId=fu-iivcCctNqGyxEiIz4aXlvS&functionSection=calls&limit=100&includeLogContext=false) (internal only, sorry).

- T=0: Container is scheduled. No queueing! Sub-second network+scheduler latency, can ignore, that matters for requests, not for cold starts.
- T=4: We print the command we're going to run. The rest of the time is container boot (1-2s?) and code in the global scope (3-2s?). We don't import `vllm` or `torch` yet.
- T=10: `vllm` logs `Automatically detected platform cuda` from `__init__.py`. So maybe five seconds to `import vllm`?
- T=18: `vllm` logs `server version` from `api_server.py`. So 7-8s to "spin up the web server"?
- T=28: `vllm` logs model/engine config (e.g. `Chunked prefill is enabled`) from `config.py`. So ~10s to read model/engine configuration?
- T=39: Over 9 seconds, `vllm` logs a `NCCL_CUMEM_ENABLE` warning, `Automatically detected platform cuda`again, and then finally `Waiting for init message from front-end` from `core.py`. Not sure what's going on here, more "engine spin-up"?
- T=45: Model is loaded (4.7s), KV cache created, model warmed up, and API server launched. Health check returns 200.

Nothing obvious jumps out to me here. Even if we loaded models via blazingly fast RDMA, we'd never save more than 5s. It would just let us run a more interesting model here, like Mistral Small, LLaMA 70B, etc.

The easiest way to speed all of this up would be to snapshot it (hard to demo in an ephemeral app, but we'd figure something out). But the created state here is going to be tricky to snapshot. Not just because it's GPU memory, but also because it involves network resources like sockets.

I investigated FP4 models, since Blackwell supports FP4 and it would cut size in ~half again. There are some testing-only versions in a neural magic side org on Hugging Face ([eg](https://huggingface.co/nm-testing/Llama-3.1-8B-Instruct-NVFP4)), and those run, but the speedup isn't meaningful and the quality seems worse. I'll keep an eye on FP4 support.

## Type of Change

## Monitoring Checklist

### Content

### DIFF
--- a/06_gpu_and_ml/llm-serving/vllm_inference.py
+++ b/06_gpu_and_ml/llm-serving/vllm_inference.py
@@ -1,5 +1,6 @@
 # ---
 # pytest: false
+# cmd: ["modal", "run", "06_gpu_and_ml/llm-serving/vllm_inference.py"]
 # ---
 
 # # Run OpenAI-compatible LLM inference with LLaMA 3.1-8B and vLLM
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d2309c0.modal.run

charlesfrye:
@prbot approve

charlesfrye:
@prbot automerge

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c5e2d04.modal.run

PR #1228, opened 2025-06-18
fix bug that occurs from wrapping q.get.aio in asyncio.wait_for

shababo:
## Bug Description
We were seeing a non-zero probability of failure for the WebRTC example. Ultimately I was able to identify the way we were using the `modal.Queue` as the culprit.

## Bug Solution
We wanted to `await q.get.aio(...)` but also have it timeout so that we could intermittently check certain conditions that would break the loop. I had wrapped it using `asyncio.wait_for`, however this seemed to cause random issues with `Queue.get` which has its own timeout and is just running a gRPC call. The solution was to use `Queue.get(timeout = ...)` instead which is more stable as it propagates the timeout to the gRPC call instead of just cancelling the task while the call is out in the ether.

Unclear to me if this is necessarily a bug in the client. `Queue.get` already has a `timeout` keyword arg so it turns out it's not necessary to use `wait_for` to get a timeout. That said, maybe we want it to be robust to task cancellation? I can make a minimal repro example and pass it on.

## Debug `print` statements
I left in some extra debug statements (though not all) that were helpful to figure out what was going on.

## Type of Change

## Monitoring Checklist

### Content

### Build Stability


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-af605fd.modal.run

charlesfrye:
lgtm

minimal repro sounds like a good idea! not obvious to me that this is a client bug, but seems worth investigating

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-225083c.modal.run

shababo:
Damn, @charlesfrye, you beat my ability to use prbot!

Agree it's not clearly a bug. That said I think it speaks to some confusion on using `modal.Queue`. Also I think how `synchronicity` plays into things is not well doc'd at all.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-1a99849.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-daf9827.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a2d3acd.modal.run

PR #1236, opened 2025-06-20
Add example to run high-throughput llm inference with tokasaurus

andrewhinh:
This is an example to run Qwen2-7B-Instruct at high throughput with Tokasaurus on 1 H200 GPU.

## Type of Change

- [☑️] New example for the GitHub repo
  - [☑️] New example for the documentation site (Linked from a discoverable page, e.g. via the sidebar in `/docs/examples`)

## Monitoring Checklist

  - [☑️] Example is configured for testing in the synthetic monitoring system, or `lambda-test: false` is provided in the example frontmatter and I have gotten approval from a maintainer
    - [☑️] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`)
    - [☑️] Example is tested by running the `cmd` with no arguments, or the `args` are provided in the example frontmatter (e.g. `args: ["--prompt", "Formula for room temperature superconductor:"]`
    - [☑️] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or other code executed locally)

### Content
  - [☑️] Example is documented with comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style
  - [☑️] All media assets for the example that are rendered in the documentation site page are retrieved from `modal-cdn.com`

### Build Stability
  - [☑️] Example pins all dependencies in container images
    - [☑️] Example pins container images to a stable tag like `v1`, not a dynamic tag like `latest`
    - [☑️] Example specifies a `python_version` for the base image, if it is used 
    - [☑️] Example pins all dependencies to at least [SemVer](https://semver.org/) minor version, `~=x.y.z` or `==x.y`, or we expect this example to work across major versions of the dependency and are committed to maintenance across those versions
      - [☑️] Example dependencies with `version < 1` are pinned to patch version, `==0.y.z`

## Outside Contributors

### DIFF
--- a/06_gpu_and_ml/llm-serving/tokasaurus_throughput.py
+++ b/06_gpu_and_ml/llm-serving/tokasaurus_throughput.py
@@ -1,0 +1,264 @@
+# ---
+# deploy: true
+# ---
+
+# # Serverless Tokasaurus (Qwen2-7B-Instruct)
+
+# In this example, we demonstrate how to use the Tokasaurus framework to serve Qwen2-7B-Instruct model
+# at very high throughput.
+
+# ## Overview
+
+# This guide is intended to document two things:
+# the general process for building Tokasaurus on Modal
+# and a specific configuration for serving the Qwen2-7B-Instruct model.
+
+# ## Set up the container image
+
+# Our first order of business is to define the environment our server will run in:
+# the [container `Image`](https://modal.com/docs/guide/custom-container).
+# Tokasaurus can be installed with `pip`, since Modal [provides the CUDA drivers](https://modal.com/docs/guide/cuda).
+
+# To take advantage of optimized kernels for CUDA 12.8, we install PyTorch, and their dependencies
+# via an `extra` Python package index.
+
+import json
+
+import aiohttp
+import modal
+
+toka_image = (
+    modal.Image.from_registry(
+        "nvidia/cuda:12.6.1-devel-ubuntu22.04", add_python="3.12"
+    )  # since tokasaurus==0.0.2 uses torch==2.6.0
+    .apt_install("git")
+    .pip_install("uv")
+    .run_commands(
+        "uv pip install --system --compile-bytecode tokasaurus==0.0.2 huggingface_hub[hf_transfer]==0.33.0"
+    )
+    .env({"HF_HUB_ENABLE_HF_TRANSFER": "1"})  # faster model transfers
+)
+
+# ## Download the model weights
+
+# We'll be running a fine-tuned instruction-following model -- Qwen2-7B-Instruct
+# that's trained to chat and follow instructions.
+
+MODEL_NAME = "Qwen/Qwen2-7B-Instruct"
+MODEL_REVISION = "f2826a00ceef68f0f2b946d945ecc0477ce4450c"  # avoid nasty surprises when repos update!
+
+# Although Tokasaurus will download weights from Hugging Face on-demand,
+# we want to cache them so we don't do it every time our server starts.
+# We'll use [Modal Volumes](https://modal.com/docs/guide/volumes) for our cache.
+# Modal Volumes are essentially a "shared disk" that all Modal Functions can access like it's a regular disk.
+
+app_name = "example-tokasaurus-throughput"
+
+hf_cache_vol = modal.Volume.from_name(f"{app_name}-hf-cache", create_if_missing=True)
+
+# ## Tuning the engine
+
+# You can tune the engine for high throughput by adjusting the following two parameters:
+# - `max_tokens_per_forward`: max tokens processed per forward pass
+# - `max_seqs_per_forward`: max sequences processed per forward pass
+
+# These two parameters work together to [maximize batch size and throughput](https://github.com/ScalingIntelligence/tokasaurus#managing-gpu-memory-with-kv-cache-limits-and-concurrency-controls).
+# Increase them jointly to find the optimal balance between memory usage and concurrency.
+# Higher values of `max_tokens_per_forward` increase throughput but use more activation memory, reducing available KV cache.
+# Higher values of `max_seqs_per_forward` increase batch size and throughput, but require larger KV cache.
+
+# Since we want to maximize the throughput, we set the batch size to the largest value we can fit in GPU RAM.
+
+MAX_BATCH_SIZE = 1024
+MAX_TOKENS_PER_FORWARD = 8192
+
+# ### Torch compile
+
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3faab6a.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b7d1720.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-262309d.modal.run

andrewhinh:
Note that at the moment, max throughput achieved is roughly 650 tok/s, much lower than ideal. This is most likely due to having to interface via the server over HTTP.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-de21bce.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d3e962a.modal.run

shababo:
Nice work! 

Summarizing the discussion @andrewhinh and I just had on how concurrent requests and auto-scaling might affect these results and our comparison to the published result.

1. It looks like the published benchmark allows the [number of concurrent requests to equal the number of items in the dataset](https://github.com/ScalingIntelligence/tokasaurus/blob/c01e494b74d29887ee3e6660b99e64c4ef170c9a/tokasaurus/benchmarks/standalone_monkeys_gsm8k.py#L176). We should confirm this.
2. We are also auto-scaling containers along with having `max_inputs = 4` for `modal.concurrency`.
  a. The published benchmark does not autoscale. We should start by setting `max_containers=1` to make sure we have concurrent inputs nailed down and get a fair comparison.
  b. Their optimization is based on caching and sharing information across requests. This means there is a tradeoff in terms of optimization gains from Tokasaurus and more containers.
  c. We want a better understanding of what GPU performance to expect and then ensure we are seeing that. The run metrics show a peak in GPU utilization then it basically goes to 0 (across all 31 containers). Maybe this is correct, but we should be sure. We should also make sure that we are maximizing GPU memory usage. [run metrics](https://modal.com/apps/modal-labs/examples/ap-1zumISHSbP5ekho8G5gqUW?start=1750828765&end=1750832172&live=false&activeTab=functions&functionId=fu-ury7xuZqDyIm7k7kwM59c1&functionSection=metrics&limit=100&includeLogContext=false)
  
If (1) is true, then the number of concurrent requests on their single H100 is 128. This is actually close to the number of concurrent requests in the run linked above (31 containers * 4 inputs). That said, we're not guaranteed to hit that max. Given that, it's a bit curious that we have so much higher throughput. It could just be that having 30 extra GPU machines means we have less queuing? 

We might want to run their code directly to see if we can repro their results.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4b0b1a8.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4eef4f5.modal.run

andrewhinh:
More thorough notes here: https://www.notion.so/modal-com/Add-e-g-for-high-throughput-llm-inference-with-Tokasaurus-21d1e7f1694980aeb63ecd9c60931cd4?source=copy_link

Tl;dr: 
- confirmed # concurrent requests = # items in dataset
- could not get working results with `max_containers=1`, presumably b/c there's an issue with how Tokasaurus queues inputs
  - edit: got working results by limiting `max_inputs`
- increasing # containers does improve throughput, but it sublinearly scales compared to linearly scaling costs
- GPU utilization does proportionally decrease with # containers, but GPU memory usage levels off after a certain # of containers
- seems if limit # GPU machines -> # inputs, queueing time ~= server startup time

charlesfrye:
Left some notes on the internal doc.

IMO we should try to beat 50k tok/s for a single replica, which puts us comfortably above sglang/vllm according to this benchmark figure from the [blogpost](https://scalingintelligence.stanford.edu/blogs/tokasaurus/):

image

That requires measuring per-replica throughput, which we're not doing here, since we measure aggregates client-side.

In my experience, this is _always_ a good idea with Modal! You want to understand the performance of individual replicas first before you add auto-scaling and multiple replicas on top.

charlesfrye (changes requested):
Screenshot 2025-06-25 at 8 20 17 PM

see above

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b8ebd50.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-fffc97e.modal.run

shababo:
Can you ensure this passes all tests and then re-ping us for review?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f2d45ad.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8de3e9c.modal.run

charlesfrye (approved):

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c9854a4.modal.run

charlesfrye:
Since our IRL discussion, I reviewed the timing logic and realized I was underestimating aggregate perf by a factor of about 2 -- we timed all requests from within the client's async task, and about 2 requests-worth of work can run in parallel for this workload.

That puts us at ~50k tok/s, which is comfortably above the performance reported for other engines on this task and is the target we set [here](https://github.com/modal-labs/modal-examples/pull/1236#issuecomment-3006905310).

Timing from the client task makes sense if each request is mapped onto a container and we want to divide out the impact of parallel execution across containers, but here we're only running one container. I used `update_autoscaler` inside of the `local_entrypoint` so that only impacts the benchmarking, not the deployment (new trick!).

I just _barely_ got the largest code block to fit in a single screen (~50 lines, from ~110) by trimming it down and moving logic out.

Screenshot 2025-07-16 at 10 33 35 PM

charlesfrye:
beep boop

Screenshot 2025-07-16 at 9 57 58 PM

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-d0ccb8f.modal.run

PR #1252, opened 2025-06-27
remove empty headers

charlesfrye:
These two examples had lines like
`# #`

which produce an empty h1 tag that 1) was likely unintended and 2) triggers an a11y warning from the Svelte compiler (below).

The preferred style for representing whitespace in the examples is actually just a blank line -- they are ignored during frontend rendering but create a similar visual separation in the raw code files.

Screenshot 2025-06-27 at 2 11 32 PM


charlesfrye:
@prbot approve

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5979b0f.modal.run

PR #1253, opened 2025-06-27
aria label for links around videos

charlesfrye:
Screen readers use link text when announcing links, but in a few places we use a video as the link content, so screen readers don't know what to announce (at least until video understanding models are a few hundred times faster).

This PR adds the appropriate [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) attributes, which resolves a warning from the Svelte Compiler.

### DIFF
--- a/06_gpu_and_ml/llm-serving/llama_cpp.py
+++ b/06_gpu_and_ml/llm-serving/llama_cpp.py
@@ -14,7 +14,7 @@
 # our jobs are still safe, for now.
 
 # <center>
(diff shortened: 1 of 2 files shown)


charlesfrye:
@prbot approve

charlesfrye:
@prbot automerge

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f2f65cc.modal.run

luiscape (commented):
✔️

wshainin:
lgtm

PR #1256, opened 2025-06-30
Add example for Kyutai STT

andrewhinh:
Add example for [Kyutai STT](https://kyutai.org/next/stt).

## Type of Change


andrewhinh:
Note this is heavily inspired by [Quillman](https://github.com/modal-labs/quillman), with some important changes:
1. Only STT, no audio playback.
2. Using [FastHTML](https://www.fastht.ml/docs/index.html) to remove the need for HTML and React.
3. Handling concurrent users in the same container.

06_gpu_and_ml/audio-to-text/kyutai/src/app.py
```diff
@@ -0,0 +1,247 @@
...
+    image=image,
+)
+@modal.concurrent(max_inputs=100)
+@modal.asgi_app()
+def web():
+    from textwrap import dedent
+
+    import fasthtml.common as fh
+
+    modal_logo_svg = dedent(
```
shababo: Can we move the web stuff (HTML, JS) to their own files. This is hard to read and makes a lot of IDE tools useless.

06_gpu_and_ml/audio-to-text/kyutai/src/common.py
```diff
@@ -0,0 +1,3 @@
+from modal import App
```
shababo: Does this need to be in its own file?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+import asyncio
+from pathlib import Path
+
+import modal
+
+from .common import app
+
+image = (
+    modal.Image.debian_slim(python_version="3.12")
+    .pip_install("uv")
```
shababo: i believe we already use `uv` under the hood. I think you can use the `pip_install` methods of `Image` here. @charlesfrye ?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+with image.imports():
+    import numpy as np
+    import sphn
+    import torch
+    from huggingface_hub import snapshot_download
+    from moshi.models import LMGen, loaders
+
+
+N_GPU = 1
+GPU_CONFIG = f"l40s:{N_GPU}"
```
shababo: Do we want to use a more powerful GPU? @charlesfrye thoughts on this?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+)
+@modal.concurrent(max_inputs=64)
+class STT:
+    @modal.enter()
+    def download_model(self):
+        print("Downloading model if necessary...")
+        snapshot_download(MODEL_NAME, revision=MODEL_REVISION)
+
+    @modal.enter()
+    def enter(self):
```
shababo: do we need two `@modal.enter` methods if we aren't using the `snap` kwarg?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+
+        web_app = FastAPI()
+
+        @web_app.get("/status")
+        async def status():
+            return Response(status_code=200)
+
+        @web_app.websocket("/ws")
+        async def websocket_endpoint(ws: WebSocket):
+            with torch.no_grad():
```
shababo: Kind of curious if performance is the same whether this context starts here or closer to the actual inference call - especially because of the async stuff.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+                await ws.accept()
+
+                # clear llm chat history + buffered audio
+                self.reset_state()
```
shababo: Maybe we only need to call this in `finally` below and not here as well?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+                                assert tokens.shape[1] == self.lm_gen.lm_model.dep_q + 1
+
+                                text_token = tokens[0, 0, 0].item()
+                                if text_token not in (0, 3):
+                                    text = self.text_tokenizer.id_to_piece(text_token)
+                                    text = text.replace("▁", " ")
+                                    msg = b"\x00" + bytes(
+                                        text, encoding="utf8"
+                                    )  # prepend "\x00" as a tag to indicate text
+                                    await ws.send_bytes(msg)
```
shababo: Putting this here makes the inference loop io-bound on this call. Honestly, it probably doesn't wait that long, but it's probably also better practice to have a third `Task` that is the `send_loop` which has a `asyncio.Queue` of inference results that it packages and sends off over the websocket.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
```diff
@@ -0,0 +1,204 @@
...
+                    await ws.close(code=1000)
+                except Exception as e:
+                    print("Exception:", e)
+                    await ws.close(code=1011)  # internal error
+                    raise e
+                finally:
+                    for task in tasks:
+                        task.cancel()
+                    await asyncio.gather(*tasks, return_exceptions=True)
+                    self.reset_state()
```
shababo: It might be the case, but just make sure this is always called when the websocket endpoint gets as far as making inferences.

shababo:
Can you also add some performance metrics? In particular, can you measure the time of each inference as well as the time to first text on the clientside?

andrewhinh:
> Can you also add some performance metrics? In particular, can you measure the time of each inference as well as the time to first text on the clientside?

Yeah, actually I took it out so I’ll just add it back in. Model loading (assuming model is downloaded) takes ~16s, TTFT is ~90ms TTFT, then ~15ms for each token after.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: This was mainly in line with [the blog post](https://kyutai.org/next/stt):

> We use this server to run Unmute; on a L40S GPU, we can serve 64 simultaneous connections at a real-time factor of 3x.

06_gpu_and_ml/audio-to-text/kyutai/src/common.py
andrewhinh: Nope, will consolidate.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: Nope, will merge.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: Isn't it necessary since we're using `modal.concurrent`, and need to reset state for each new ws connection?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: Tested, when it's moved right before `chunk = torch.from_numpy(chunk)`, latency for each token seems to go down by ~1ms, but it's not consistent so hard to tell if it's by chance.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: Ah interesting, didn't catch that. Though you're also right in that latency doesn't go down with this fix.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: Could you elaborate?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: The case for this is made clearer in [54e2c6e](https://github.com/modal-labs/modal-examples/pull/1256/commits/54e2c6e28cd2a1b300fd03a5cc369411821e3cb9).

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
shababo: Unclear to me that you can measure an effect size that small :)

I suppose it's best to make this scope as small as possible without requiring entering and exiting the context more than once.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
shababo: Hmm... I'm not sure what exactly the state is here. If this is the case, we should be careful we're not doing something wrong. You're clearing the state so it doesn't interfere with the new connection/client, yes? If yes, then aren't you also killing the previous client's state?

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
shababo: Ignore this, the previous thread about state/concurrency will cover what should happen here.

06_gpu_and_ml/audio-to-text/kyutai/src/stt.py
andrewhinh: I'm assuming since max_inputs is set, there would be no overwrites, though I'll add an assertion.

shababo:
Sorry if I wasn't clear before. For time to first text on the client side, we should start measuring from the time user starts speaking to the time we get back the first chunk audio. Right now it looks like it starts measure the time as soon as the page is loaded.

Session started
TTFT: 34366.48ms <----- THIS IS NOT THE CORRECT MEASUREMENT
Latency: 20.56ms
Latency: 19.91ms
Latency: 19.69ms
Latency: 19.97ms
Latency: 19.96ms
Latency: 19.68ms
Latency: 18.39ms
Latency: 18.47ms
Latency: 20.0ms

shababo:
I'd like @charlesfrye to confirm he's down with this, but my take is that we should remove the effort to implement concurrency. I don't have time to debug it and the behavior I'm seeing is that it's not all that close to working.

@andrewhinh if you remove that, then I believe it should be close? do you need to add explanatory copy? what else is left do if we abandon concurrency?

charlesfrye:
1. We don't need explanatory copy here -- just working code for now.
2. We can ship this with just a single user per container first.

06_gpu_and_ml/llm-serving/ollama.py
shababo, line 1: Why are these changes in this PR?

06_gpu_and_ml/audio-to-text/kyutai/app.py
```diff
@@ -0,0 +1,362 @@
...
+
+        self.moshi = checkpoint_info.get_moshi(device=self.device)
+        self.lm_gen = LMGen(
+            self.moshi,
+            # sampling params
+            temp=0,
+            temp_text=0,
+        )
+
+        self.batch_size = 1
```
shababo: Let's make this a constant of the class.

06_gpu_and_ml/audio-to-text/kyutai/frontend/modal-logo.svg
shababo, line 1: Replace this with a logo that uses the correct font.

06_gpu_and_ml/llm-serving/ollama.py
shababo: Oh snap. It was my fault. Removed!

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f241f26.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-74d1855.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-8227105.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4ec16b4.modal.run

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,608 @@
...
+        self.rtf = 1000
+
+    @modal.web_server(port=PORT, startup_timeout=10 * MINUTES)
+    def web_server(self):
+        import os
+        import subprocess
+
+        config_path = hf_cache_vol_path / "config-stt-en_fr-hf.toml"
+        config_content = dedent(
+            """static_dir = "./static/"
```
shababo: Can this toml be in its own file instead of inline?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,608 @@
...
+            max_seq_len = 40960
+
+            [modules.asr.model.extra_heads]
+            num_heads = 4
+            dim = 6
+            """
+        )
+        if not config_path.exists():
+            with open(config_path, "w") as f:
+                f.write(config_content)
```
shababo: Wait... why are we writing a static file inside of our code? This speaks to just making a .toml file. What was the motivation for this?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,608 @@
...
+
+        env = os.environ.copy()
+        env["LD_LIBRARY_PATH"] = LD_LIBRARY_PATH
+        env["CUDA_HOME"] = CUDA_HOME
+        env["CANDLE_CUDA_COMPUTE_CAP"] = CUDA_COMPUTE_CAP
```
shababo: Aren't we setting some of this in the build definition already?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,608 @@
...
+
+            opus_stream_inbound = sphn.OpusStreamReader(self.sample_rate)
+            transcription_queue = asyncio.Queue()
+            is_first_token = True
+            stream_start_time = None
+            last_audio_receive_time = None
+
+            print("Session started")
+
+            async with websockets.connect(
```
shababo: Think about what this is doing. You are opening a websocket connection where both endpoints are on the same machine.

The solve here is that we should have two versions, not one. I'll expand on this in my global review comment.

shababo (commented):
Ok, my apologies if I wasn't clear on this yesterday. But we want two versions. We still want the previous app that didn't use the Rust server. And we want a version that simply deploys the Rust server.

Also, you sort of mixed the two together. You only need the Modal websocket endpoint in the non-server case. There should only ever be a client <> container websocket. There is no need for another. 

Also, make sure you understand how your app is distributed. Right now, you have a  websocket connection that is between two processes on the same machine.

shababo:
Oh, sometimes I get a DriverError. e.g. [this run](https://modal.com/apps/modal-labs/shababo-dev/ap-RBwCZ96jBxX7zUEB7JKIQi?start=1752089375.436&end=1752092975.436&live=true&activeTab=functions&functionId=fu-mIjUDAJGp7HUNnmB0ByPPr&containerId=ta-01JZRD2HYQKMKPK2QMWHBXEKE8&functionSection=containers&limit=100&includeLogContext=false)

andrewhinh:
> Oh, sometimes I get a DriverError. e.g. [this run](https://modal.com/apps/modal-labs/shababo-dev/ap-RBwCZ96jBxX7zUEB7JKIQi?start=1752089375.436&end=1752092975.436&live=true&activeTab=functions&functionId=fu-mIjUDAJGp7HUNnmB0ByPPr&containerId=ta-01JZRD2HYQKMKPK2QMWHBXEKE8&functionSection=containers&limit=100&includeLogContext=false)

Yeah saw this too, will look into it.

andrewhinh:
> Ok, my apologies if I wasn't clear on this yesterday. But we want two versions. We still want the previous app that didn't use the Rust server. And we want a version that simply deploys the Rust server.
> 
> You sort of mixed the two together. You only need the Modal websocket endpoint in the non-server case. There should only ever be a client <> container websocket. There is no need for another.
> 
> Also, make sure you understand how your app is distributed. Right now, you have a websocket connection that is between two processes on the same machine.
> 
> Also also, please make sure you address all comments, even the nitty ones.

Sorry about that, will address all.

shababo:
> > Oh, sometimes I get a DriverError. e.g. [this run](https://modal.com/apps/modal-labs/shababo-dev/ap-RBwCZ96jBxX7zUEB7JKIQi?start=1752089375.436&end=1752092975.436&live=true&activeTab=functions&functionId=fu-mIjUDAJGp7HUNnmB0ByPPr&containerId=ta-01JZRD2HYQKMKPK2QMWHBXEKE8&functionSection=containers&limit=100&includeLogContext=false)
> 
> Yeah saw this too, will look into it.

If you saw this, then don't request a review!!! Fix it or get help fixing it! I already mentioned that this sort of thing wastes the team's time. I literally spent time making sure I wasn't doing anything wrong or hadn't inadvertently changed the code before reporting. I checked to see if it occurred with both `modal run` and `modal serve`, etc...

Please do not submit PRs for review unless you think they could feasibly be approved.

Also, to be super clear. Asking for help, asking questions, learning - these are all excellent behaviors. It is much preferred over quickly delivering many incomplete projects on your own.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Agree with creating a separate toml file. I did this because `moshi-server` requires the config be passed as a local file as far as I know.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Yeah, I remember I set it here also and it seemed to fix cuda driver mismatch errors, but maybe I'm wrong. Probably since you said you still get the error.

andrewhinh: Nope, just needed to pass os.environ, sorry!

andrewhinh:
> > > Oh, sometimes I get a DriverError. e.g. [this run](https://modal.com/apps/modal-labs/shababo-dev/ap-RBwCZ96jBxX7zUEB7JKIQi?start=1752089375.436&end=1752092975.436&live=true&activeTab=functions&functionId=fu-mIjUDAJGp7HUNnmB0ByPPr&containerId=ta-01JZRD2HYQKMKPK2QMWHBXEKE8&functionSection=containers&limit=100&includeLogContext=false)
> > 
> > 
> > Yeah saw this too, will look into it.
> 
> If you saw this, then don't request a review!!! Fix it or get help fixing it! I already mentioned that this sort of thing wastes the team's time. I literally spent time making sure I wasn't doing anything wrong or hadn't inadvertently changed the code before reporting. I checked to see if it occurred with both `modal run` and `modal serve`, etc...
> 
> Please do not submit PRs for review unless you think they could feasibly be approved.
> 
> Also, to be super clear. Asking for help, asking questions, learning - these are all excellent behaviors. It is much preferred over quickly delivering many incomplete projects on your own.

Sorry about this, still unlearning how I used to do things at other companies. I appreciate the direct feedback. I've separated but not pushed the rust example as I can't get it to work consistently, but the python frontend and local entrypoint consistently work.

charlesfrye:
> still unlearning how I used to do things at other companies

Yeah, this is a somewhat surprising thing about Modal's PR culture! Still working on helping folks onboard to it properly here in the `modal-examples` repo.

shababo:
> > still unlearning how I used to do things at other companies
> 
> Yeah, this is a somewhat surprising thing about Modal's PR culture! Still working on helping folks onboard to it properly here in the `modal-examples` repo.

Just want to make sure one of the distinctions is clear here: it's not just about short-lived PRs-as-final-filter vs. long-lived PRs-as-project-discussion-forum, it's also about having known bugs in your PR that reviewers end up re-discovering. Even in the long-lived case it would be good to at least call these out so review can address them (or not) more efficiently.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-967e99e.modal.run

06_gpu_and_ml/kyutai/app.py
```diff
@@ -20,9 +21,10 @@
     modal.Image.debian_slim(python_version="3.12")
     .pip_install("uv")
     .run_commands(
-        "uv pip install --system --compile-bytecode moshi==0.2.6 fastapi==0.115.14 hf_transfer==0.1.9",
+        "uv pip install --system --compile-bytecode moshi==0.2.9 fastapi==0.115.14 hf_transfer==0.1.9 julius==0.2.7",
     )
     .env({"HF_HUB_ENABLE_HF_TRANSFER": "1"})
+    .add_local_file(Path(__file__).parent / "bria.mp3", "/root/bria.mp3")
```
shababo: Please put this on the CDN so that it's not dependent on users having this file locally. Alternatively, is this file available publicly as URL?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+    .pip_install("uv")
+    .run_commands(
+        "uv pip install --system --compile-bytecode moshi==0.2.9 fastapi==0.115.14 hf_transfer==0.1.9 julius==0.2.7",
+    )
+    .env({"HF_HUB_ENABLE_HF_TRANSFER": "1"})
+    .add_local_file(Path(__file__).parent / "bria.mp3", "/root/bria.mp3")
+)
+
+MODEL_NAME = "kyutai/stt-1b-en_fr"
+MODEL_REVISION = "40b03403247f4adc9b664bc1cbdff78a82d31085"
```
shababo: Do we need a specific revision for the model?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+
+
+@app.function(
+    scaledown_window=600,
+    timeout=600,
+    image=web_image,
+)
+@modal.concurrent(max_inputs=100)
+@modal.asgi_app()
+def web():
```
shababo: Let's move this function to the bottom of the file. It's very hard to parse and is not the focus of the example. Better to focus the readers on what's important.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6e7f447.modal.run

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+hf_cache_vol_path = Path("/root/.cache/huggingface")
+volumes = {hf_cache_vol_path: hf_cache_vol}
+
+
+@app.function(
+    scaledown_window=600,
+    timeout=600,
+    image=web_image,
+)
+@modal.concurrent(max_inputs=100)
```
shababo: Why do we need concurrent inputs for this `cls`? Or if not necessary, what are we gaining?

The source code for the clientside is obfuscated so hard to say, but it could even break things?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
+import asyncio
+import base64
+import dataclasses
+import time
+from pathlib import Path
+
+import modal
+
+app = modal.App(name="kyutai")
```
shababo: nit: please change to something slightly more descriptive - even `kyutai-streaming-stt` would be great.

06_gpu_and_ml/kyutai/app.py
shababo: And in general, we should avoid adding binary files to the repo. This one is probably small and not the end of the world, but if there's another option, it's usually better.

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+MODEL_NAME = "kyutai/stt-1b-en_fr"
+MODEL_REVISION = "40b03403247f4adc9b664bc1cbdff78a82d31085"
+
+hf_cache_vol = modal.Volume.from_name(f"{app.name}-hf-cache", create_if_missing=True)
+hf_cache_vol_path = Path("/root/.cache/huggingface")
+volumes = {hf_cache_vol_path: hf_cache_vol}
+
+
+@app.function(
+    scaledown_window=600,
```
shababo: Why do we need a long scaledown window for this function?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+                ),
+                fh.Script(app_js),
+                cls="relative bg-gray-900 text-white min-h-screen flex flex-col items-center justify-center p-4",
+            ),
+        )
+
+    return fast_app
+
+
+MINUTES = 60
```
shababo, line 25: This should also be used earlier when you are setting durations.

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def __str__(self):
+        return f"{self.text} ({self.timestamp[0]:.2f}:{self.timestamp[1]:.2f})"
+
+
+@app.cls(
+    image=stt_image,
+    gpu="l40s:1",
+    volumes=volumes,
+    scaledown_window=15 * MINUTES,
```
shababo: Why do we want this scaledown window?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+        )
+        self.audio_delay_seconds = checkpoint_info.stt_config.get(
+            "audio_delay_seconds", 5.0
+        )
+        self.padding_token_id = checkpoint_info.raw_config.get(
+            "text_padding_token_id", 3
+        )
+
+        # warmup gpus
+        for _ in range(4):
```
shababo, line 78: Does it matter if we warm this up on actual audio vs zeros?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+                    if not isinstance(data, bytes):
+                        print("received non-bytes message")
+                        continue
+                    if len(data) == 0:
+                        print("received empty message")
+                        continue
+
+                    if stream_start_time is None:
+                        stream_start_time = time.monotonic_ns()
+                    last_audio_receive_time = time.monotonic_ns()
```
shababo: Put time-marking methods as close to the moment you want to mark as possible. E.g.,
```Python
while True:
    data = await ws.receive_bytes()
    received_time = time.monotonic_ns()
    if not isinstance(data, bytes):
         print("received non-bytes message")
         continue
    if len(data) == 0:
        print("received empty message")
        continue
    last_audio_receive_time = received_time
```

Not a huge issue here - especially because you really want to be doing the timing on the clientside - both is fine but clientside is more relevant.

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+                    transcription_queue, \
+                    is_first_token, \
+                    stream_start_time, \
+                    last_audio_receive_time
+                all_pcm_data = None
+
+                while True:
+                    await asyncio.sleep(0.001)
+
+                    pcm = opus_stream_inbound.read_pcm()
```
shababo, line 204: Is it possible wait on `read_pcm`? Then we can lose the `await asyncio.sleep(...)` line?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+                    while all_pcm_data.shape[-1] >= self.frame_size:
+                        if is_first_token and stream_start_time is not None:
+                            ttft_ms = round(
+                                (time.monotonic_ns() - stream_start_time) / 1e6, 2
+                            )
+                            is_first_token = False
+                            stream_start_time = None
+                            print(f"TTFT: {ttft_ms}ms")
+
+                        chunk = all_pcm_data[: self.frame_size]
```
shababo: We should use the model's built in VAD here. And generally it looks like you based this more on the Quillman example than the Kyutai inference code. [See here](https://github.com/kyutai-labs/delayed-streams-modeling/blob/baf0c75bba89608e921cb26e03c959981df2ad5f/scripts/stt_from_file_pytorch.py#L181).

Can we make sure we are using this model optimally/using the patterns provider by the developers?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+            async def send_loop():
+                """
+                Reads outbound data, and sends it across websocket
+                """
+                nonlocal transcription_queue
+                while True:
+                    await asyncio.sleep(0.001)
+
+                    try:
+                        text = transcription_queue.get_nowait()
```
shababo: We don't need `get_nowait()`... we can just wait, right? Or am I missing something?

06_gpu_and_ml/kyutai/app.py
```diff
@@ -0,0 +1,561 @@
...
+                text_tokens.shape[-1], last_segment_start + self.mimi.frame_rate
+            )
+        else:
+            last_segment_end = last_segment_start + end_of_last_segment[0]
+        _decode_segment(int(last_segment_start), int(last_segment_end))
+
+        return sequence_timestamps
+
+    @modal.method()
+    def process_audio_file(self, audio_file: str):
```
shababo: Instead of having this here which uses different code than the actual app - and therefore making it not great for testing the app, let's use a file and "stream" it to the app like we do in the [Parakeet test](https://github.com/modal-labs/modal-examples/blob/8521e53ddf7019691a61d72a8bdce33c04601e97/06_gpu_and_ml/audio-to-text/parakeet.py#L291).

If there's a reason that won't work, let me know.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Yep, here: https://github.com/kyutai-labs/delayed-streams-modeling/raw/refs/heads/main/audio/bria.mp3

06_gpu_and_ml/kyutai/app.py
andrewhinh: I suppose not, though I've always done it as a good habit.

06_gpu_and_ml/kyutai/app.py
andrewhinh: I guess it's a habit of mine to give cheaper containers longer scaledown windows since it's not expensive to keep them running for longer, but it's not strictly necessary.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Nope, see the [kyutai rust eg](https://github.com/kyutai-labs/delayed-streams-modeling/blob/main/scripts/stt_from_file_rust_server.py#L69) where they do the same.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Not strictly necessary, but as noted [here](https://modal.com/docs/guide/webhooks#websockets), it's safe and beneficial since each container can manage multiple websockets without interfering with each other and spinning up new containers.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Oops, yeah you're right.

06_gpu_and_ml/kyutai/app.py
andrewhinh: I don't think [sphn](https://github.com/kyutai-labs/sphn) has an async interface for that method.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Yeah I guess it's not too important if container startup times are under 15s.

06_gpu_and_ml/kyutai/app.py
andrewhinh: Not strictly necessary, especially since container startup time is pretty low.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-12b33f8.modal.run

andrewhinh:
All comments addressed!

06_gpu_and_ml/kyutai/app.py
shababo: Copy that. Not a big deal, just wondering. It's generally not common practice, but I'm not against.

06_gpu_and_ml/kyutai/app.py
shababo: Ok, great. As long we are sure it works and there isn't any weird state sharing.

andrewhinh: Yep, confirmed via testing. Conceptually, each client concurrently connects to `web()` and get served the page via the same container, but each client connects to the STT class via its own websocket connection, so state can't be shared between clients.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-fc9247a.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-184bd46.modal.run

PR #1257, opened 2025-07-01
Switch flux kontext to fp8

andrewhinh:
Switch flux kontext to run in fp8.

## Type of Change

### DIFF
--- a/06_gpu_and_ml/stable_diffusion/image_to_image.py
+++ b/06_gpu_and_ml/stable_diffusion/image_to_image.py
@@ -121,8 +121,7 @@ class Model:
             text_encoder_2=text_encoder_2.to(self.device),
             torch_dtype=dtype,
             cache_dir=CACHE_DIR,
-            device_map="balanced",
-        )
+        ).to(self.device)
 
     @modal.method()
     def inference(


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9035881.modal.run

06_gpu_and_ml/stable_diffusion/image_to_image.py
```diff
@@ -90,25 +90,59 @@ class Model:
     @modal.enter()
     def enter(self):
         print(f"Downloading {MODEL_NAME} if necessary...")
+
+        dtype = torch.bfloat16
+
+        self.seed = 42
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        print(f"Using device: {self.device}")
```
luiscape: Is there a point on adding this if the define defined in the function is a GPU? Despite that, would it be useful to run this model on a CPU?

06_gpu_and_ml/stable_diffusion/image_to_image.py
```diff
@@ -90,25 +90,59 @@ class Model:
...
-            torch_dtype=torch.bfloat16,
-            device_map="balanced",
+            revision=MODEL_REVISION,
+            transformer=transformer,
+            text_encoder_2=text_encoder_2,
+            torch_dtype=dtype,
             cache_dir=CACHE_DIR,
         )
 
+        self.pipe.enable_model_cpu_offload()
```
luiscape: Why is this needed?

06_gpu_and_ml/stable_diffusion/image_to_image.py
```diff
@@ -90,25 +90,59 @@ class Model:
...
+
+        self.seed = 42
+        self.device = "cuda" if torch.cuda.is_available() else "cpu"
+        print(f"Using device: {self.device}")
+
+        transformer = FluxTransformer2DModel.from_single_file(
+            "https://huggingface.co/Comfy-Org/flux1-kontext-dev_ComfyUI/blob/main/split_files/diffusion_models/flux1-dev-kontext_fp8_scaled.safetensors",
+            torch_dtype=dtype,
+            cache_dir=CACHE_DIR,
+        )
```
luiscape, line 103: Doesn't this mean that the weights will be downloaded on every cold boot? It would be better to create a new function that performs these operations then saves the quantized model to a Modal Volume.

06_gpu_and_ml/stable_diffusion/image_to_image.py
andrewhinh: Not necessarily, it’s more a fallback in case the user were to remove the `gpu=`.

06_gpu_and_ml/stable_diffusion/image_to_image.py
andrewhinh: No, since cache_dir (which points to a Modal Volume) is specified and will be checked first.

06_gpu_and_ml/stable_diffusion/image_to_image.py
luiscape: Is this useful, though? Would this realistically run if there's no GPU?

andrewhinh: Fair enough, will remove.

06_gpu_and_ml/stable_diffusion/image_to_image.py
andrewhinh: Oops forgot to remove, was initially getting CUDA_HOME errors hence the switch to the Nvidia cuda image, but forgot to use device_map. Will fix

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-af4b468.modal.run

06_gpu_and_ml/stable_diffusion/image_to_image.py
```diff
@@ -90,25 +90,57 @@ class Model:
...
             MODEL_NAME,
-            revision="f9fdd1a95e0dfd7653cb0966cda2486745122695",
-            torch_dtype=torch.bfloat16,
-            device_map="balanced",
+            revision=MODEL_REVISION,
+            transformer=transformer.to(self.device),
+            text_encoder_2=text_encoder_2.to(self.device),
+            torch_dtype=dtype,
             cache_dir=CACHE_DIR,
+            device_map="balanced",
```
luiscape: I am struggling to understand if this is good or not. My understanding is that this will split the model evenly between GPUs. But in this case we only have one GPU. Why do we need to set this?

andrewhinh: Yeah you're right, using `.to()` is simpler.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-34a5767.modal.run

PR #1265, opened 2025-07-07
Add example for running a remote (stateless) MCP server on Modal

ehdr:
It demonstrates how to deploy a simple [MCP server](https://modelcontextprotocol.io/) on Modal, that provides a tool to get the current date and time in a given timezone. It is a stateless MCP server, meaning that it does not store any state between requests, and uses the "streamable HTTP" transport type.

It uses the [FastMCP library](https://github.com/jlowin/fastmcp) to create the MCP server, which is wrapped using FastAPI, and served with a [Modal web endpoint](https://modal.com/docs/guide/webhooks).

image

## Type of Change

## Monitoring Checklist

### Content
  - ~[ ] All media assets for the example that are rendered in the documentation site page are retrieved from `modal-cdn.com`~

### Build Stability

## Outside Contributors

### DIFF
--- a/misc/mcp_server_stateless.py
+++ b/misc/mcp_server_stateless.py
@@ -24,46 +24,52 @@ image = modal.Image.debian_slim(python_version="3.12").pip_install(
     "fastmcp~=2.10.2",
 )
 
-with image.imports():
-    from fastapi import FastAPI
+
+# First, we create the MCP server itself using FastMCP, and add a tool to it that
+# allows LLMs to get the current date and time in a given timezone.
+def make_mcp_server():
     from fastmcp import FastMCP
 
-# First, we create the MCP server itself, using FastMCP.
-mcp = FastMCP("Date and Time MCP Server")
+    mcp = FastMCP("Date and Time MCP Server")
 
+    @mcp.tool()
+    async def current_date_and_time(timezone: str = "UTC") -> str:
+        """Get the current date and time.
 
-# Next, we add a tool to the MCP server, that allows LLMs to get the current date and
-# time in a given timezone.
-@mcp.tool()
-async def current_date_and_time(timezone: str = "UTC") -> str:
-    """Get the current date and time.
+        Args:
+            timezone: The timezone to get the date and time in (optional). Defaults to UTC.
 
-    Args:
-        timezone: The timezone to get the date and time in (optional). Defaults to UTC.
+        Returns:
+            The current date and time in the given timezone, in ISO 8601 format.
+        """
+        from datetime import datetime
+        from zoneinfo import ZoneInfo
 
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-632288b.modal.run

misc/mcp_server_stateless.py
```diff
@@ -0,0 +1,102 @@
...
+    return datetime.now(tz).isoformat()
+
+
+# We then use FastMCP to create a Starlette app with `streamable-http` as transport
+# type, and set `stateless_http=True` to make it stateless. This will then be mounted by
+# FastAPI.
+mcp_app = mcp.http_app(transport="streamable-http", stateless_http=True)
+
+fastapi_app = FastAPI(lifespan=mcp_app.router.lifespan_context)
+fastapi_app.mount("/", mcp_app, "mcp")
```
aksh-at: This shouldn't be in global scope right? Because then the user needs `fastmcp` and `fastapi` installed locally. It does make writing the doc text harder though.

aksh-at (approved):
Lgtm besides comment!

misc/mcp_server_stateless.py
charlesfrye: In case it makes the structuring easier: the current rule for examples is that you can have fastapi imports and other objects in the global scope, since we so often need that for the `fastapi_endpoint` feature. Still have to move `fastmcp` and `@mcp` though.

ehdr: Thanks, pushed a change to move them out now!

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-aa1e925.modal.run

misc/mcp_server_stateless.py
```diff
@@ -0,0 +1,108 @@
...
+    fastapi_app.mount("/", mcp_app, "mcp")
+
+    return fastapi_app
+
+
+# Finally we deploy the FastAPI app as a Modal web endpoint, using [the `asgi_app`
+# decorator](https://modal.com/docs/reference/modal.asgi_app):
+@app.function(image=image)
+@modal.asgi_app()
+def web_endpoint():
```
charlesfrye, line 79: Hmm, why not just make this function return the FastAPI app?

IMO splitting them into two functions is more confusing than clarifying. We want to express to users that they can just take whatever function previously built their FastAPI app and wrap it with `asgi_app` to deploy to Modal.

ehdr: Yes, makes total sense. Fixing in #1271!

PR #1267, opened 2025-07-07
Update Tailscale example to use `modal.is_local()`

luiscape:
This guarantees that a configuration expected to happen only inside the container instead of applying them with `image.imports()`. This is better because imports might work locally, leading to the application of container-bound configurations locally.

### DIFF
--- a/10_integrations/tailscale/modal_tailscale.py
+++ b/10_integrations/tailscale/modal_tailscale.py
@@ -27,12 +27,15 @@ image = (
 )
 app = modal.App(image=image)
 
-# Configure Python to use the SOCKS5 proxy globally.
+# Packages might not be installed locally. This catches import errors and
+# only attemps imports in the container. 
 with image.imports():
     import socket
 
     import socks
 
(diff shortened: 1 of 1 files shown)


luiscape:
@prbot approve @aksh-at

aksh-at (approved):

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5d55390.modal.run

10_integrations/tailscale/modal_tailscale.py
```diff
@@ -27,12 +27,15 @@
 )
 app = modal.App(image=image)
 
-# Configure Python to use the SOCKS5 proxy globally.
+# Packages might not be installed locally. This catches import errors and
+# only attemps imports in the container.
```
charlesfrye:
```suggestion
# only attempts imports in the container.
```

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-977025f.modal.run

PR #1280, opened 2025-07-16
[DRAFT] add Sandbox test case generator example

zhang-lucy:
## Type of Change

## Monitoring Checklist

    - [ ] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`)
    - [ ] Example is tested by running the `cmd` with no arguments, or the `args` are provided in the example frontmatter (e.g. `args: ["--prompt", "Formula for room temperature superconductor:"]`
    - [ ] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or other code executed locally)

### Content
  - [ ] Example is documented with comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style

### Build Stability
  - [ ] Example pins all dependencies in container images
    - [ ] Example pins container images to a stable tag like `v1`, not a dynamic tag like `latest`
    - [ ] Example pins all dependencies to at least [SemVer](https://semver.org/) minor version, `~=x.y.z` or `==x.y`, or we expect this example to work across major versions of the dependency and are committed to maintenance across those versions

## Outside Contributors


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-e2ef8ed.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-4379dbb.modal.run

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
+# ---
```
zhang-lucy, line 1: Would love feedback on high level structure before diving too much into the weeds - please see PR description for high level q's.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-ac554e9.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-18fcb12.modal.run

charlesfrye:
> My plan is to get this code fully passing tests and in a good state, then add in content, comments, etc. in a follow-up PR. Let me know if this sounds good

Yes, definitely!

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        }
+    )
+    .entrypoint([])  # silence noisy logs
+)
+
+with model_image.imports() as imports:
+    from transformers import AutoModelForCausalLM, AutoTokenizer
+
+
+# GH Repo Configs
```
charlesfrye, line 32: Can this config be passed from a command-line interface?

Otherwise this example requires code changes to change the target repo + branch.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        with open(f"/data/outputs/{output_file_name}", "w") as f:
+            f.write(output_contents)
+        print(f"Output written to {output_file_name}")
+        return output_file_name
+
+    @modal.method()
+    def generate(self, file_name: str) -> str:
+        file_contents, test_file_contents = self.load_inputs(file_name)
+
+        system_prompt = get_system_prompt()
```
charlesfrye, line 84: Are these two components really fixed? I would expect some parts to be variable and so better to pass in via a command-line interface or similar.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        run_sandbox(output_file)
+
+
+# # Addenda
+# The below functions are utility functions.
+
+
+def get_user_prompt(file_text: str, test_file_text: str) -> str:
+    return f"""
+    You are an expert Python test engineer. Your task is to improve an existing test file using `pytest`.
```
charlesfrye, line 216: I think the "you are an expert" prompts are no longer needed -- something of an artifact of the earlier models still being too close to a pure language model.

But rather than trusting me, probably better to go off of a prompting guide for the model, assuming one is available, or suggestions from folks online (r/LocalLlama loves/loved DeepSeek Coder). And if we can link out to those resources, that's an even better thing to share with users.

charlesfrye: Cf https://boomerprompts.com

charlesfrye: Relatedly, there's a lot of overlap between the system and user prompts -- not sure how you're deciding what goes where.

The distinction is pretty hard to define when the users are generally trusted. Otherwise, the system prompt is mostly intended to restrain the LLM's behavior.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+    return (
+        "You are a senior software engineer with expertise in test-driven development and Python unit testing. "
+        "You write clean, idiomatic unit tests using the `pytest` framework, based on source code with functions or classes. "
+        "Your task is to enhance an existing test file by adding more test cases. Focus on edge cases, input validation, and untested behavior, especially as inferred from docstrings and type hints. "
+        "Do not change or add import statements. Do not explain your reasoning. Output only a complete, valid Python file. "
+        "Limit each line to a maximum of 100 characters to avoid output truncation or formatting errors."
+    )
+
+
+def post_process(output: str) -> str:
```
charlesfrye, line 251: Would be preferable to use structured outputs/function calling/JSONSchema here -- which is I think much more work coming from `transformers` and probably a reason to use SGLang or vLLM.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+    },
+    gpu="L40S",
+)
+class Deepseek:
+    @modal.enter()
+    def load_model(self):
+        MODEL_NAME = "deepseek-ai/deepseek-coder-6.7b-instruct"
+        REVISION = "e5d64addd26a6a1db0f9b863abf6ee3141936807"
+        self.tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME, revision=REVISION)
+        self.model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, revision=REVISION)
```
charlesfrye, line 53: I think it might actually be just as easy, and likely more performant, to use vLLM or SGLang here. But not via the OpenAI-compatible API, which adds more complexity (e.g. HTTP server and client) -- just via their `LLM` Python classes.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+        file_contents, test_file_contents = self.load_inputs(file_name)
+
+        system_prompt = get_system_prompt()
+        user_prompt = get_user_prompt(file_contents, test_file_contents)
+
+        messages = [
+            {"role": "system", "content": system_prompt},
+            {"role": "user", "content": user_prompt},
+        ]
+        print("Applying chat template...")
```
charlesfrye, line 91: I'd skip this one -- it's a bit too inside-baseball to be worth sharing and it's fast.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+            messages, add_generation_prompt=True, return_tensors="pt"
+        ).to("cuda")
+        print("Generating...")
+        outputs = self.model.generate(
+            inputs,
+            max_new_tokens=1024,
+            do_sample=False,
+            num_return_sequences=1,
+        )
+        print("Decoding...")
```
charlesfrye, line 102: I'd skip this one -- it's a bit too inside-baseball to be worth sharing and it's fast.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,274 @@
...
+
+@app.cls(
+    image=model_image,
+    volumes={
+        "/cache": model_volume,
+        "/data": files_volume,
+    },
+    gpu="L40S",
+)
+class Deepseek:
```
charlesfrye, line 47: Naming nit that IMO is actually pretty deep -- this isn't really a `DeepSeek` class. It's more like a `DeepSeekPythonTestWriter` or a `PythonTestWriter`.

Maybe that's how you can think of what goes in the system prompt: the system prompt directs the model to generally behave as indicated by the class name. User prompts describe the specific behavior/provide the relevant context for it to serve their needs.

charlesfrye:
> What should the overall conclusion of this example be? The model generally outputs a couple of failing test files and a couple of passing ones. Failing is not a bad thing; hopefully the generated test point to failing edge cases. However, this is not an ideal end state for an example (i.e. showing users error messages)

I don't have a full answer here, but I have two guiding questions:
- what would a user want to do with the failures?
- how do engineers normally look at test results?

For the former: they probably want to look at the test code and the trace. Can we share that with them, directly or via a pointer? By pointer, I mean a URL or a file path on a Volume. If the latter, we'd want to include a CLI command for downloading it.

For the latter: can we share a test report using some Python package or tool? Nothing immediately comes to mind, because I'm a bit behind on the tooling there.

> For our N test files generated, should there be 1 Sandbox or N sandboxes to evaluate? Multiple Sandboxes makes it easier to pinpoint which test file is failing, but also adds complexity.

Modal's big win is horizontal scaling, so IMO N Sandboxes is better. We can take that N up to 1k concurrently without breaking a sweat; we run 10k+ for users in production. That's impressive and it's what we'd want to call out in the text and promotion for the example. It's also probably what a customer we share the example with is trying to build.

It's furthermore part of our brand that horizontal scaling is ergonomic, not just possible. So if scaling to many Sandboxes makes the example less than delightful, I think the move would be to 1) flag the disergonomicity internally, to spark a discussion and possible surface better patterns and then 2) rewrite the example once better patterns are available. Being able to link to this example as a concrete example of a disergonomicity (in e.g. a Linear ticket) is hugely beneficial for eng.

> Any additional comments on structure very welcome, as well as on the [dummy repo](https://github.com/modal-labs/password-analyzer)

See code review. The other thing I'd suggest is executing a `local_entrypoint` instead of a remote Function. I find that makes it easier to separate concerns that are about managing local resources or communicating with the user (e.g. printing a URL, as described above) from concerns that are about doing the "real work" on Modal infra.

Additionally, you're using a persisted Volume for everything. That's a kinda spooky amount of shared state across runs. Can we use an ephemeral Volume somewhere or otherwise cleanly break up state? Didn't think too hard on this one, just wanted to flag a gut reaction.

zhang-lucy:
Thanks @charlesfrye for the detailed comments! Closing as suggested for longer PR iteration cycle.

PR #1287, opened 2025-07-21
Boltz fix: use input msa file instead of msa server due to rate limit

harmya:
Context: https://modal-com.slack.com/archives/C044YUDC1QU/p1753116951664599?thread_ts=1753056102.982579&cid=C044YUDC1QU

We were getting rate limits from the external msa server for supplying sequence alignments. Used the example msa in the boltz2 repo as a file and passing its contents to the remote function. File is small enough to be able to do this.

## Type of Change

## Monitoring Checklist

## Outside Contributors


06_gpu_and_ml/protein-folding/boltz_predict.py
```diff
@@ -121,20 +130,21 @@ def main(
...
+    msa_path.write_text(input_msa)
+
     args = shlex.split(args)
 
     print(f"🧬 predicting structure using boltz model from {models_dir}")
     subprocess.run(
-        ["boltz", "predict", input_path, "--use_msa_server", "--cache", str(models_dir)]
-        + args,
-        check=True,
+        ["boltz", "predict", input_path, "--cache", str(models_dir)] + args,
```
charlesfrye: why'd we drop the `check=True` here?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6449b8f.modal.run

06_gpu_and_ml/protein-folding/boltz_predict.py
harmya: Fixed, was iterating, put it back

06_gpu_and_ml/protein-folding/boltz_predict.py
```diff
@@ -121,20 +130,21 @@ def main(
...
 )
-def boltz_inference(boltz_input_yaml: str, args="") -> bytes:
+def boltz_inference(boltz_input_yaml: str, input_msa: str, args: str = "") -> bytes:
     import shlex
     import subprocess
 
     input_path = Path("input.yaml")
     input_path.write_text(boltz_input_yaml)
 
+    msa_path = Path("seq1.a3m")
```
charlesfrye: This will only work if this filename is hard-coded in the YAML file.

That means it will fail if the input YAML has more than one protein in it, e.g. when trying to do structure prediction for interacting proteins. That's a pretty big use case for Boltz.

We could get around this by mounting a local directory so that the YAML file and any associated files are all pulled in.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-a983ca6.modal.run

06_gpu_and_ml/protein-folding/boltz_predict.py
```diff
@@ -90,10 +79,18 @@ def main(
...
 
+# Since protein folding requires an input yaml file and an input msa file, we add the "data" directory in our local directory to the image.
+
 # Here, we do it in a few lines, using the `uv` package manager for extra speed.
 
-image = modal.Image.debian_slim(python_version="3.12").run_commands(
-    "uv pip install --system --compile-bytecode boltz==2.1.1"
+data_dir = (
+    here / "data"
+)  # Use the 'here' variable which is defined as Path(__file__).parent
```
charlesfrye, line 88: What does the comment add here?

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-cded9b4.modal.run

06_gpu_and_ml/protein-folding/boltz_predict.py
```diff
@@ -90,10 +79,18 @@ def main(
 # Because Modal images include [GPU drivers](https://modal.com/docs/guide/cuda) by default,
 # installation of higher-level packages like `boltz` that require GPUs is painless.
 
+# Since protein folding requires an input yaml file and an input msa file, we add the "data" directory in our local directory to the image.
```
charlesfrye, line 82: 1. They don't just need an input yaml file and an input msa file. They may need a number of files. We're simplifying this by saying "well, just make sure all the files are in the `data` directory".
2. I would split the explanation in two parts. First, separate the Image definition into two steps -- installing the package and then adding the data directory. Explain that we mount the local data directory so that it's accessible to the remote code. Elsewhere, when explaining how to use this script, note that the input YAML files and any files it references, e.g. MSA files, need to be in that directory.

06_gpu_and_ml/protein-folding/data/boltz_affinity.yaml
```diff
@@ -3,7 +3,7 @@ sequences:
   - protein:
       id: A
       sequence: MVTPEGNVSLVDESLLVGVTDEDRAVRSAHQFYERLIGLWAPAVMEAAHELGVFAALAEAPADSGELARRLDCDARAMRVLLDALYAYDVIDRIHDTNGFRYLLSAEARECLLPGTLFSLVGKFMHDINVAWPAWRNLAEVVRHGARDTSGAESPNGIAQEDYESLVGGINFWAPPIVTTLSRKLRASGRSGDATASVLDVGCGTGLYSQLLLREFPRWTATGLDVERIATLANAQALRLGVEERFATRAGDFWRGGWGTGYDLVLFANIFHLQTPASAVRLMRHAAACLAPDGLVAVVDQIVDADREPKTPQDRFALLFAASMTNTGGGDAYTFQEYEEWFTAAGLQRIETLDTPMHRILLARRATEPSAVPEGQASENLYFQ
-      msa: ./seq1.a3m
+      msa: data/msa/seq1.a3m
```
charlesfrye, line 6: This path should stay relative to the YAML file

06_gpu_and_ml/protein-folding/boltz_predict.py
harmya: Will update.

harmya:
Closing this for now, will open another one with the changes. Will have to figure out testing with this change.

PR #1288, opened 2025-07-22
Reorganize STT examples, resurrect streaming whisper

charlesfrye:
This PR organizes our speech-to-text examples:
- Streaming and batched whisper
- Streaming Kyutai
- Streaming Parakeet

They are all placed into a single folder. The file names and App names for the examples are normalized. Functions named "web" are renamed to be more clear about function (API or UI). The phrase "real-time" is dropped, in favor of "streaming".

The streaming whisper example was quite old and no longer worked. It was nested deeply enough that it wasn't being tested. Biggest problem was that the PyTube project is no longer updated and the requests for audio were blocked. I switched over to just raw audio files.

The "Charlie Chaplin Dictator Speech" wasn't a great demo for streaming Whisper responses, since with 30s chunks and the default settings, the ~330s file is split into only two chunks of a few hundred seconds. I ended up going with a podcast episode, the first of _The History of Rome_ by Mike Duncan. I put it on the Modal CDN (after cutting out the ad at the beginning). This file is split into ~20 chunks of ~30s each. I also added cacheing to a Volume so that the model weights don't have to be downloaded each time. That should end up looking pretty good at first run on a 10 GPU limit (on the Starter Plan).

Bigger issue: the 30s window size makes Whisper a bad model for "streaming" anyway. The other examples are better. I'm keeping the example around because the code in it is interesting (very different from the other streaming examples) and because the model is still reasonably popular.


charlesfrye:
@prbot approve

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-50e4ac2.modal.run

PR #1289, opened 2025-07-22
delete duplicate parakeet example file

charlesfrye:
Removes duplicate parakeet example file, introduced in #1288 to allow us to rename to `streaming_parakeet` following a PR to the monorepo


luiscape (approved):

charlesfrye:
This should fail on the Preview Docs job, because it ran on `824308d`, a commit from before the rename+redirect was merged. I will then re-run the job once the rename+redirect is merged to verify that the pair of PRs has the intended effect.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3f2272e.modal.run

PR #1290, opened 2025-07-22
[CI] Fix support for workflow dispatch, for running changed example

zhang-lucy:
[Before](https://github.com/modal-labs/modal-examples/actions/runs/16451074809) - Action is workflow dispatch-able, but the diff matrix is not enabled.
[After](https://github.com/modal-labs/modal-examples/actions/runs/16451745289) - Action is workflow dispatch-able!

Also couple of light changes to the Action names so they're a bit clearer.

💡 Idea: In the spirit of increasing workflow dispatches, it would be nice to have a "driver" GH action that can be dispatchable outside of a PR.

## Type of Change

### DIFF
--- a/.github/workflows/build-and-run-example.yml
+++ b/.github/workflows/build-and-run-example.yml
@@ -1,4 +1,4 @@
-name: Build and run example
+name: Build and run random example
 
 on:
   workflow_dispatch:
--- a/.github/workflows/run-examples.yml
+++ b/.github/workflows/run-examples.yml
@@ -1,4 +1,4 @@
-name: Run
+name: Run changed examples
 
 on:
   pull_request:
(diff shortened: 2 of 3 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-cabe474.modal.run

.github/workflows/build-and-run-example.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and run example
+name: Run random example
```
charlesfrye: We'd want to keep the "Build" here, since the main difference between this workflow and the changed example workflow is that it also builds the image by ignoring cached images. We use this to track regressions in builds.

charlesfrye (approved):
Nice! Love making it possible to manually trigger workflows.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-95eedd7.modal.run

PR #1295, opened 2025-07-24
install pytorch with cuda before trtllm to avoid cpu-only version

mmangkad:
## Type of Change

## Summary
Reorder pip install commands to install PyTorch with CUDA before TensorRT-LLM to prevent CPU-only PyTorch installation.

When TensorRT-LLM is installed first, it pulls in PyTorch (2.7.1) as a dependency but defaults to the CPU-only version. By installing PyTorch with CUDA support first, we ensure the correct CUDA-enabled version is used instead of falling back to the CPU variant.

### DIFF
--- a/misc/trtllm_deepseek.py
+++ b/misc/trtllm_deepseek.py
@@ -55,10 +55,10 @@ tensorrt_image = (
     )
     .pip_install("uv")
     .run_commands(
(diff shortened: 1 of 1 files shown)


charlesfrye (approved):
thanks!

PR #1296, opened 2025-07-28
Add Sandbox test case generator example - 2nd pass

zhang-lucy:
⚡ Changes since last PR (https://github.com/modal-labs/modal-examples/pull/1280)
- Prompt Engineering - Tried a variety of different techniques - surprisingly,  the repetitive hints seem to be required for keeping the code working. I'll keep iterating a bit more but would prefer to keep this relatively error-free
- Sandbox Bells-and-Whistles: Added 2 tunnels - one for showing the code diff, one for [Allure](https://allurereport.org/) - cute testing report framework. See screenshots!
- SGLang: Made it the client-server model as [recommended](https://modal-com.slack.com/archives/C08G83NB9B3/p1753288465186799).
- Various cleanups + responded to comments 

‼️ Outstanding Issues
- Error is thrown with the final `asyncio.gather`, as the `.wait()` function throws an error on timeout. Will be fixed in subsequent PR
- Volumes should be ephemeral

💡 Outstanding Q's
- Overall Runtime: ~3:20. Can we do better? The server startup time is the most obvious place to cut down on. Taking any recs!
    - Image build ~60 sec
    - Server startup time ~75 sec
    - Inference ~45 sec
    - Sandboxes ~20 sec
- Deepseek Coder-6.7B-Instruct? Originally I picked this model since it was small and seemed to do pretty well on [the leaderboard](https://huggingface.co/spaces/bigcode/bigcode-models-leaderboard). Could try a bigger model if we're doing client-server.
- How much retry + additional validation logic should we add? The current generate function will error if the length of the output exceeds the set limit - this is flaky as it's really prompt driven
- Edit: Just realized I didn't add the ephemeral volumes piece. Will do in a subsequent push.

❓ General high level question is just how much optimization to do + what's missing here? The focus of this example is _Sandboxes_; we have plenty of inference examples elsewhere.

image
image
image
image

## Type of Change

## Monitoring Checklist

### Build Stability

## Outside Contributors

### DIFF
--- a/13_sandboxes/test_case_generator.py
+++ b/13_sandboxes/test_case_generator.py
@@ -1,7 +1,10 @@
 # ---
 # cmd: ["modal", "run", "-m", "13_sandboxes.test_case_generator"]
-# args: ["--gh-owner", "modal-labs", "--gh-repo-name", "password-analyzer", "--gh-module-path", "src/password_strength", "--gh-test-dir-path", "tests", "--gh-branch", "main"]
+# args: ["--gh-owner", "modal-labs", "--gh-repo-name", "password-analyzer", "--gh-module-path", "src/password_strength", "--gh-tests-path", "tests", "--gh-branch", "main"]
 # ---
+import subprocess
+import time
+
 import modal
 
 app = modal.App(
@@ -10,14 +13,19 @@ app = modal.App(
 model_volume = modal.Volume.from_name("deepseek-model-volume", create_if_missing=True)
 files_volume = modal.Volume.from_name("files-volume", create_if_missing=True)
 
+MODEL_NAME = "deepseek-ai/deepseek-coder-6.7b-instruct"
+MODEL_REVISION = "e5d64addd26a6a1db0f9b863abf6ee3141936807"
+
 
 model_image = (
-    modal.Image.from_registry("nvidia/cuda:12.4.0-devel-ubuntu22.04", add_python="3.12")
-    .pip_install(
+    modal.Image.from_registry("lmsysorg/sglang:v0.4.9.post3-cu126", add_python="3.12")
+    .uv_pip_install(
+        "sglang[all]==0.4.9.post3",
         "transformers==4.53.2",
         "torch==2.7.1",
         "accelerate==1.8.1",
         "hf_transfer==0.1.9",
+        "numpy<2",
     )
     .env(
         {
@@ -28,16 +36,6 @@ model_image = (
     .entrypoint([])  # silence noisy logs
 )
 
-with model_image.imports() as imports:
-    from transformers import AutoModelForCausalLM, AutoTokenizer
-
-
-# GH Repo Configs
-# GH_OWNER = "modal-labs"
-# GH_REPO_NAME = "password-analyzer"
-# GH_MODULE_NAME = "password_strength"
-# GH_BRANCH = "main"
-
 
 @app.cls(
     image=model_image,
@@ -47,14 +45,59 @@ with model_image.imports() as imports:
     },
     gpu="L40S",
 )
-class TestFileGenerator:
+@modal.concurrent(max_inputs=3)  # Each container runs up to 3 requests at once.
+class TestCaseServer:
     @modal.enter()
-    def load_model(self):
-        MODEL_NAME = "deepseek-ai/deepseek-coder-6.7b-instruct"
-        REVISION = "e5d64addd26a6a1db0f9b863abf6ee3141936807"
-        self.tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME, revision=REVISION)
-        self.model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, revision=REVISION)
-        self.model.to("cuda")
+    def download_model(self):
+        from huggingface_hub import snapshot_download
+
+        snapshot_download(
+            MODEL_NAME,
+            local_dir=f"/cache/{MODEL_NAME}",
+            revision=MODEL_REVISION,
+            ignore_patterns=["*.pt", "*.bin"],
+        )
+
+    @modal.enter()
+    def start_model_server(self):
+        import subprocess
+
+        serve_params = {
+            "host": "0.0.0.0",
+            "port": 8000,
+            "model": f"/cache/{MODEL_NAME}",
+            "log-level": "error",
+        }
+        serve_cmd = "python -m sglang.launch_server " + " ".join(
+            [f"--{k} {v}" for k, v in serve_params.items()]
+        )
+
+        self.serve_process = subprocess.Popen(serve_cmd, shell=True)
+        wait_for_port(self.serve_process, 8000)
+
+        print("SGLang server is ready!")
+
+    @modal.web_server(port=8000, startup_timeout=240)
+    def serve(self):
+        return
+
+    @modal.method()
+    def healthcheck(self):
+        import requests
+
+        return requests.get(f"{self.serve.get_web_url()}/health").status_code == 200
+
+
+@app.cls(
+    image=modal.Image.debian_slim(python_version="3.12").uv_pip_install(
+        "openai==1.97.1"
+    ),
+    volumes={
+        "/data": files_volume,
+    },
+)
+class TestCaseClient:
+    url: str = modal.parameter()
 
     def load_inputs(self, file_name: str) -> tuple[str, str]:
         import os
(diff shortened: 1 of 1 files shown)


13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+        wait_for_port(self.serve_process, 8000)
+
+        print("SGLang server is ready!")
+
+    @modal.web_server(port=8000, startup_timeout=240)
+    def serve(self):
+        return
+
+    @modal.method()
+    def healthcheck(self):
```
zhang-lucy: this isn't called but feels right to have

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+            response_format={
+                "type": "json_schema",
+                "json_schema": {
+                    "name": "test_file",
+                    "schema": json_schema,
+                },
+            },
+        )
+        output = response.choices[0].message.content
+        try:
```
zhang-lucy: open to add'l suggestions on error handling and/or retries

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+    sb = modal.Sandbox.create(
+        "sh",
+        "-c",
+        cmd,
+        app=app,
+        image=image,
+        volumes={
+            "/data": files_volume,
+        },
+        timeout=300,  # TODO: should we keep this as the default?
```
zhang-lucy: will remove

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+        tunnel1 = sb.tunnels()[8000]
+        tunnel2 = sb.tunnels()[8001]
+        print(f"Sandbox created and run for generated test file: {filename}")
+        print(f"✨ View diff: {tunnel2.url}")
+        print(f"✨ View test results: {tunnel1.url}\n")
+
+    return file_to_sandbox.values()
+
+
+def poll_sandboxes(sandboxes: list[modal.Sandbox]):
```
zhang-lucy: hacky; keeps the app alive so it doesn't kill all the sandboxes

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+# # Addenda
+# The below functions are utility functions.
+def create_sandboxes(filenames: list[str], gh_owner: str, gh_repo_name: str):
+    file_to_sandbox: dict[str, modal.Sandbox] = {}
+    for filename in filenames:
+        print(f"Running sandbox for {filename}")
+        image = get_sandbox_image(gh_owner, gh_repo_name)
+        sb = run_sandbox(image, filename)
+        file_to_sandbox[filename] = sb
+    time.sleep(20)
```
zhang-lucy: hack to make sure that the tunnels show up at the very end, after the noisy sandbox logs. don't think there's a sane way to silence those logs, but i feel like it's ok? kinda cool to see pytest running in the background in the sandboxes.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,379 @@
...
+    3. Based on that understanding, **add** new test cases to the test file to increase coverage—especially edge cases, boundary conditions, and untested branches.
+    4. Use `pytest` idioms, but do **not** add or change import statements—**use only what is already imported**.
+    5. Do **not** explain your reasoning—just return the final modified test file.
+
+    ### Requirements:
+    - Your output must be a valid, complete Python file with the added test cases.
+    - Do not modify existing test logic unless necessary to support your new test cases.
+    - Do not import any additional modules.
+    - Limit each line to a maximum of 100 characters to avoid output truncation or formatting errors.
+    - Limit your output to around 25 lines. Make sure to complete any functions or blocks you start.
```
zhang-lucy, line 325: this particular line needs to be present in both the system and user prompt to minimize flakiness. will experiment with what other lines i can remove/de-dupe.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,380 @@
...
+model_volume = modal.Volume.from_name("deepseek-model-volume", create_if_missing=True)
+files_volume = modal.Volume.from_name("files-volume", create_if_missing=True)
+
+MODEL_NAME = "deepseek-ai/deepseek-coder-6.7b-instruct"
+MODEL_REVISION = "e5d64addd26a6a1db0f9b863abf6ee3141936807"
+
+
+model_image = (
+    modal.Image.from_registry("lmsysorg/sglang:v0.4.9.post3-cu126", add_python="3.12")
+    .uv_pip_install(
```
pawalt, line 22: Why do all these deps need to be installed if we're using the sglang image?

13_sandboxes/test_case_generator.py
pawalt: I don't think we should have it if it's not used. It's the job of some calling application to health check anyways.

13_sandboxes/test_case_generator.py
```diff
@@ -0,0 +1,380 @@
...
+)
+@modal.concurrent(max_inputs=3)  # Each container runs up to 3 requests at once.
+class TestCaseServer:
+    @modal.enter()
+    def download_model(self):
+        from huggingface_hub import snapshot_download
+
+        snapshot_download(
+            MODEL_NAME,
+            local_dir=f"/cache/{MODEL_NAME}",
```
pawalt: Does `local_dir` work like `cache_dir`?

pawalt (commented):
I think this needs to be written in the literate programming style - this is just code right now.

zhang-lucy:
> I think this needs to be written in the literate programming style - this is just code right now.

@pawalt this PR will be just code - will add in the narrative text in a follow-up PR

13_sandboxes/test_case_generator.py
zhang-lucy: removed some, kept the ones for hf transfer

13_sandboxes/test_case_generator.py
zhang-lucy: yup! https://huggingface.co/docs/huggingface_hub/en/guides/download#download-files-to-a-local-folder

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-6dc4477.modal.run

charlesfrye:
> I think this needs to be written in the literate programming style - this is just code right now.

Quick note here -- I'd actually prefer folks write new examples with only code + comments the first time around, then we can do the full literate programming/documentation site treatment afterwards before/when we promote the example to become discoverable.

13_sandboxes/test_case_generator.py
pawalt: Can you add a comment to this effect?

13_sandboxes/test_case_generator.py
pawalt: Can you just raise here? Or is it too noisy? If it's too noisy, I think you should add a 5 second, 10 retry system. We shouldn't swallow errors.

13_sandboxes/test_case_generator.py
pawalt: I also don't have a good suggestion here. Can you add a `HACK` comment?

13_sandboxes/test_case_generator.py
pawalt: Can you just `asynctio.join` on `sb.wait.aio`?

13_sandboxes/test_case_generator.py
zhang-lucy: Changed it to just raise here - I think structured inputs should be fine

13_sandboxes/test_case_generator.py
zhang-lucy: Added

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-496205a.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-1ac2004.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-b63d469.modal.run

zhang-lucy:
This example passes all checks except for the final line:

```python
await asyncio.gather(*[sb.wait.aio(raise_on_termination=False) for sb in sandboxes])
```

because `.wait()` default throws an error (`raise_on_termination=False` doesn't prevent the error being raised on timeout). Merging because there's been enough iteration on this example; will fix this in a follow-up PR. Also updated description.

PR #1299, opened 2025-07-29
Add example on using a stateful sandbox executor

thomasjpfan:
This example showcases how to build a `SandboxExecutor` to create a stateful sandbox code executor. The public API for this executor looks like:

```python
executor = SandboxExecutor(app_name="jupyter-example")
codes = ["a = 10", "a", "print(a + 100)", "a / 0"]
for code in codes:
    output = executor.run_code(code)
    print(output)
executor.terminate()
```

By using a Jupyter kernel, we can distinguish between the result of an expression and the output of `stdout` and `stderr`.

For this example, it requires the [websockets](https://pypi.org/project/websockets/) to connect to the Jupyter kernel. The simple way to run the script is `uv run sandbox_executor.py`.

## Type of Change

## Monitoring Checklist

    - [ ] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`)
    - [ ] Example is tested by running the `cmd` with no arguments, or the `args` are provided in the example frontmatter (e.g. `args: ["--prompt", "Formula for room temperature superconductor:"]`
    - [ ] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or other code executed locally)

### Content

### Build Stability
  - [ ] Example pins all dependencies in container images
    - [ ] Example pins container images to a stable tag like `v1`, not a dynamic tag like `latest`
    - [ ] Example pins all dependencies to at least [SemVer](https://semver.org/) minor version, `~=x.y.z` or `==x.y`, or we expect this example to work across major versions of the dependency and are committed to maintenance across those versions

### DIFF
--- a/misc/sandbox_executor.py
+++ b/misc/sandbox_executor.py
@@ -17,8 +17,8 @@
 # values from the `SandboxExecutor`.
 
 import json
-import secrets
 import re
+import secrets
 import uuid
 from dataclasses import dataclass, field
 from http.client import RemoteDisconnected


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-31d9b64.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-20cf290.modal.run

misc/sandbox_executor.py
```diff
@@ -0,0 +1,229 @@
...
+    result: Optional[str] = None
+    logs: list[str] = field(default_factory=list)
+    error: Optional[ExecutionError] = None
+
+
+# The `SandboxExecutor` creates a Modal sandbox and starts a Jupyter Kernel gateway to
+# run code in a Jupyter kernel.
+
+
+@dataclass
```
charlesfrye, line 53: Non-blocking, just curious: why is this a `dataclass`? Maybe I'm behind on Pythonisms, but I thought `dataclass`es were intended to be used as record types, and that it was better etiquette to make a full class if you're going to add methods.

charlesfrye (approved):
slick!

misc/sandbox_executor.py
thomasjpfan: I see `dataclasses` as full classes where I do not need to write as much boilerplate. Some features I like:

1. Adding member variables with types automatically generates a `__init__`
2. A nice `__repr__` out of the box
3. `dataclass(frozen=True)`
4. Stuff like `mylist: list = field(default_factory=list)` means I do not need to write:

```python
class SandboxExecutor:
    def __init__(self, mylist: list | None = None):
        self.mylist = [] if mylist is None else mylist
```

> it was better etiquette to make a full class if you're going to add methods.

I have been okay with adding methods to dataclasses. In the [Python dataclasses docs](https://docs.python.org/3/library/dataclasses.html), the first example is a dataclass with a method.

charlesfrye: Nice! I'll have to try that pattern out.

PR #1316, opened 2025-08-02
remove trtllm_throughput deploy and drop dbt_modal_inference

charlesfrye:
We previously considered making more "chained" examples that consume other examples as services. The only case we ever built was the `dbt_modal_inference` demo, which consume the high-throughput LLM inference service provided by `trtllm_throughput`.

We're backing off from that idea now, in favor of building demos and applications outside of this repo to demonstrate/explore patterns more complex than what fits in a single example.

The `dbt_modal_inference` demo was cool, but didn't get much love, internally or externally, and the creator is no longer at the company, so we don't have a good maintainer.


charlesfrye:
@prbot approve

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-c872667.modal.run

PR #1324, opened 2025-08-06
Add SGLang server implementation for GPT OSS

WayneLe17:
Added an example of running GPT-OSS (20B BF16) with [sglang](https://github.com/sgl-project/sglang) on Modal using the H100 GPU. This example demonstrates how to set up the environment, configure and load the model, and expose an ASGI app for serving inference.


charlesfrye:
Thanks for sharing!

PR #1331, opened 2025-08-13
Fix outlines example

zhang-lucy:
Old version of `outlines` was dependent on `pyairports`, which no longer exists

## Type of Change

- [x] Example updates (Bug fix)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3fc6431.modal.run

PR #1332, opened 2025-08-13
update outlines version and add pydantic

zhang-lucy:
This example was breaking due to an outdated `outlines` version, which relied on `pycountry`, a package that no longer exists. Also updated the schema to use Pydantic.

## Type of Change


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-592cd2d.modal.run

shababo (approved):
lgtm! thanks for catching and fixing this!

PR #1335, opened 2025-08-13
Fix AttributeError in ColQwen2 Chat with PDF Vision Example

zhang-lucy:
## Type of Change

### DIFF
--- a/06_gpu_and_ml/llm-serving/chat_with_pdf_vision.py
+++ b/06_gpu_and_ml/llm-serving/chat_with_pdf_vision.py
@@ -36,7 +36,9 @@ model_image = (
     .apt_install("git")
     .pip_install(
         [
-            "git+https://github.com/illuin-tech/colpali.git@782edcd50108d1842d154730ad3ce72476a2d17d",  # we pin the commit id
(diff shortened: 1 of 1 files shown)


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f5cd54b.modal.run

06_gpu_and_ml/llm-serving/chat_with_pdf_vision.py
```diff
@@ -36,7 +36,9 @@
     .apt_install("git")
     .pip_install(
         [
-            "git+https://github.com/illuin-tech/colpali.git@782edcd50108d1842d154730ad3ce72476a2d17d",  # we pin the commit id
+            "colpali-engine==0.3.5",
+            "transformers>=4.45.0",
```
charlesfrye, line 40: Could we pin`transformers` and `torch` more tightly? An `~=` should do the trick.

You might also find the versions currently used in a recent Image in the `examples` environment and then pin with `~=` to that patch -- in case this is resolved to something higher.

PR #1337, opened 2025-08-15
update gpt-oss to latest revision

mmangkad:
update gpt-oss to latest revision re chat template fixes

## Type of Change


andrewhinh:
@charlesfrye checked with `modal run` and seems good.

PR #1339, opened 2025-08-18
Fix verifiers by using uv pip install

harmya:
Was getting a `pip._vendor.packaging.requirements.InvalidRequirement: Parse error at "'(>=0.24.'": Expected string_end` error (see run [here](https://github.com/modal-labs/modal-examples/actions/runs/17024473802)) which could be solved by upgrading pip however used uv_pip_install instead to fix this + speed up the installation. **No other changes.**

## Type of Change


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-5a8ef4a.modal.run

PR #1344, opened 2025-08-27
add an example for NSight Systems profiling

charlesfrye:
This PR adds a simple demonstration of how to run the NSight Systems profiler on Modal.

It primarily documents the installation process and basic integration patterns (direct Function output, Volume, `shell`).

## Type of Change

## Monitoring Checklist

### Content

### Build Stability


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-0fa93b1.modal.run

charlesfrye:
@prbot approve

06_gpu_and_ml/nsight-profiling/nsys.py
```diff
@@ -0,0 +1,126 @@
...
+# compile it with `nvcc`, and then profile it with `nsys`.
+
+# We both return the generated profile to the caller
+# and persist the profile to a [Modal Volume](https://modal.com/docs/guide/volumes).
+
+profile_volume = modal.Volume.from_name("example-nsys-traces", create_if_missing=True)
+
+
+@app.function(image=image, gpu="A10", volumes={"/traces": profile_volume})
+def compile_and_profile(code: str, output_path: str = "profile.nsys-rep"):
```
shababo, line 69: Is it a bad idea to provide this type of `eval` interface at the same time as using `runc` which allows additional syscalls?

I actually have no idea, but it jumps out as me as a possibly bad thing to do. But I'd also believe that it's guaranteed safe... I suppose it's not fundementally different from us running their code that's contained in a file...

internal/utils.py
```diff
@@ -35,6 +35,7 @@ class Example(BaseModel):
     stem: Optional[str] = None  # stem of path
     tags: Optional[list[str]] = None  # metadata tags for the example
     env: Optional[dict[str, str]] = None  # environment variables for the example
+    runtimes: Optional[list[str]] = None  # list of Modal Function runtimes
```
shababo, line 38: Why is this a list? I looked at the rest of the code and unless i'm missing something we don't run the example once for each runtime do we?

Seems like we use `runc` if it's in there, otherwise `gvisor`...?

shababo (commented):
Seems like this runs properly but just had a few questions - potentially just for my own clarification.

internal/utils.py
charlesfrye: It's a list to match the interface in benchmark synthetic monitoring. More for keeping the impedance between the two systems low than for any good reason

06_gpu_and_ml/nsight-profiling/nsys.py
charlesfrye: This is intended for development use -- you wouldn't deploy this Function as an endpoint.

PR #1350, opened 2025-09-03
Refactored dependency installation

ShivamB25:
**Latest commits:**
1. `2e7c752` - "Refactor dependency installation to use uv_pip_install across multiple files"
2. `8c6745b` - "Refactor code for improved readability and consistency in various scripts {black .}"

## Summary

Refactored dependency installation across the entire codebase to use [`uv_pip_install()`](https://modal.com/docs/reference/modal.Image#uv_pip_install) instead of [`pip_install()`](https://modal.com/docs/reference/modal.Image#pip_install) for improved build performance. Additionally applied code formatting improvements using Black formatter for better code consistency.

The changes include:
- **Performance improvement**: Switched from `pip_install()` to `uv_pip_install()` across 80+ example files, which provides significantly faster package installation during container builds
- **Code formatting**: Applied Black formatter to improve code readability and consistency across various scripts

### DIFF
--- a/02_building_containers/import_sklearn.py
+++ b/02_building_containers/import_sklearn.py
@@ -13,7 +13,9 @@ import modal
 
 app = modal.App(
     "example-import-sklearn",
-    image=modal.Image.debian_slim().apt_install("libgomp1").pip_install("scikit-learn"),
+    image=modal.Image.debian_slim()
+    .apt_install("libgomp1")
+    .uv_pip_install("scikit-learn"),
 )
 
 # The `app.image.imports()` lets us conditionally import in the global scope.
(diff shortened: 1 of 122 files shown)


ShivamB25:
@charlesfrye i have tested for mostly and all of them are working as intended

charlesfrye:
@ShivamB25 thanks for the work, but I can't accept this PR.

1. The `uv_pip_install` feature is still new and we'd rather not use it everywhere. We've noted some impedance mismatches between `uv` and containerized environments, e.g. the strong push to use a virtual environment and the need to opt into bytecode compilation, so we want to go slowly, rather than immediately recommending it as THE way to install Python packages in Modal Images.
2. We use `ruff`, not `black`.

PR #1351, opened 2025-09-03
Add error handling for test case gen

zhang-lucy:
<!--
  ✍️ Write a short summary of your work. Screenshots and videos are welcome!
-->

## Type of Change

Test case generator example was being [flaky](https://modal-com.slack.com/archives/C044YUDC1QU/p1756857683821389). Further [inspection](https://github.com/user-attachments/files/22124146/output.txt) showed that the following line was throwing a JSONDecodeError ~25-50% of the time.

```python
output_contents = json.loads(output)["file_contents"]                                                                                                                                                                                                                               
```

This is due to the underlying model not outputting valid JSON, on occasion. Not ideal of course, but adding a try-catch. Could also catch the JSONDecodeError instead, but I'd also like to catch any errors where the JSON is valid but the expected `file_contents` key is not outputted. And any other LLM nonsense.

### DIFF
--- a/13_sandboxes/test_case_generator.py
+++ b/13_sandboxes/test_case_generator.py
@@ -152,8 +152,12 @@ class TestCaseClient:
             },
         )
         output = response.choices[0].message.content
-        output_contents = json.loads(output)["file_contents"]
-        return self.write_outputs(f"test_{file_name}", output_contents)
+        try:
+            output_contents = json.loads(output)["file_contents"]
+            return self.write_outputs(f"test_{file_name}", output_contents)
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


charlesfrye (approved):
We could probably do better on error-handling here -- enforce JSON on the LLM with grammar or retries and/or bubble exceptions up from here and handle them at the client (e.g. ignoring in `local_entrypoint` invocations) -- but this is better than what we have.

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9c989f4.modal.run

PR #1371, opened 2025-10-10
restrict accepted encodings

charlesfrye:
zstd compression was [recently added to aiohttp](https://docs.aiohttp.org/en/stable/changes.html), in version 3.13.0 on 2025-10-06. The next day, we started picking up failures in this example in our synthetic monitoring system with the following error message:

```
ClientPayloadError: 400, message: Can not decode content-encoding: zstd
```

It appears that in some environments, installing aiohttp 3.13.0 does not also install dependencies for zstd decompression (and it's not in the stdlib before 3.14).

This seems like an aiohttp bug. But it's not worth investigating further, since we can work around it here by explicitly opting out of zstd as an encoding.

## Type of Change


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-69dd395.modal.run

charlesfrye:
@prbot approve

shababo:
lgtm!

PR #1373, opened 2025-10-15
remove the musicgen example

charlesfrye:
This PR removes our example of Meta's MusicGen.

The image builds for MusicGen started failing two days ago (cf synthetic monitors). The error message indicates an issue when building the PyAV package. Didn't root cause, but the package had a release around the same time. The MusicGen model is run via the audiocraft package, which is fairly research-y code, so we should expect build problems to accumulate over time.

#1372 added a `generate_music` example that uses a more recent model, so let's just use that one.


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-55367c5.modal.run

charlesfrye:
@prbot approve

shababo:
lgtm

PR #1380, opened 2025-10-30
Hunyuan Gamecraft example

klyap:
Hunyuan GameCraft world model lets you generate consistent, first-person style videos by providing:
- a starting image
- a text prompt
- a sequence of actions (e.g., move forward/backward, turn left/right)

The app produces a video simulating movement within the generated world!

- This example uses the distilled model checkpoint.
- SageAttention / FP8
    - Hunyuan Gamecraft repo supports SageAttention by using the `—use_sage=True` flag, but this was finicky to set up on Modal, so skipping this speed optimization. 
- More GPUs
    - Each generation of a ~1 second video should take ~300 seconds on 1 H100.
    - Tried running on 8xH100’s but it didn’t help much with speeding up
- Volume download pattern from https://github.com/modal-labs/modal-examples/blob/main/misc/lmdeploy_oai_compatible.py

## Type of Change

### Content

### Build Stability

## Outside Contributors

https://github.com/user-attachments/assets/301adf66-58e5-4b31-bdcc-562272445ca9

### DIFF
--- a/misc/hunyuan_gamecraft.py
+++ b/misc/hunyuan_gamecraft.py
@@ -3,19 +3,19 @@
 # This script demonstrates how to deploy the Hunyuan GameCraft interactive web UI on Modal.
 #
 # ## Overview
-# First, download the model weights to the Modal volume.
+# First, download the model weights to the Modal Volume. This can take up to an hour.
 # ```bash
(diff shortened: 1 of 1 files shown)


charlesfrye:
thanks!

PR #1383, opened 2025-10-31
improved getting started/demo examples

charlesfrye:
**We're tightly focused on AI infrastructure now, so let's show off how our infrastructure works for AI in our demos.**

This PR updates the existing inference example to use an LLM instead of an STT model. STT is not AI, but LLMs might be.

It also adds a number of examples demonstrating extensions of that base example: batch processing with `.map` to show easy scaling, serving with `fastapi_endpoint` to show easy serving, and perf optimization with Volumes and Memory Snapshots to show performance features. The final example brings all of the pieces together in a single script.

The added examples are all in `01_getting_started`. Like the other examples in that folder, they are intended to support live demos, rather than act as documentation, so they are not commented.

## Base: `inference.py`

This is the "base" example, showing bare-bones execution of an LLM on Modal. It is as svelte as I could make it, at 25 lines and ~700 chars.

The model is Qwen 3 1.7B in FP8 quant. The model was selected to balance intelligence & speed and because it is ungated -- no Secrets required. The end-to-end latency is on the order of 30s. We print the prompt to reduce the apparent "latency" to users. I couldn't figure out how to get HF to consistently print the progress bars while loading a `Pipeline`, so there's an unfortunate ~15s of quiescence when the model is loaded.

The example is "self-documenting": if you `modal run` it with no arguments, it passes its own code directly to the model and asks for an explanation. The explanations I have seen during testing are reasonable, if not perfect, as shown below.

Screenshot 2025-10-30 at 5 13 03 PM

## Three simple extensions: `inference_map.py`, `inference_perf.py`, and `inference_endpoint.py`

The three extensions are intended to support demos to audiences interested in different aspects of the Modal platform. Each is also a pretty good demo for showing the dashboard.

- `inference_map.py` runs the LLM in parallel across every Python file in the repo -- nearly 100 files. This is probably the fastest "time-to-value" demo, because it viscerally demonstrates our rapid scale-up capabilities. Running on dozens of H100s is both cool and a flex. It is close kin to the classic "`square` plus `nvidia-smi` on many H100s" demo. It's good in general, but it's best for backend ML types. It takes about a minute to run -- make sure to point out the 100x scaleup but only 2x latency increase.

Screenshot 2025-10-30 at 5 54 22 PM

- `inference_endpoint.py` adds a `fastapi_endpoint`. It's a nice addition to the base demo to show that we're good for deployment, not just one-off data processing scripts. Hit the URL directly to get responses in the browser or go to `/docs` and then hit the endpoint for something more visually appealing. The Swagger/OpenAPI docs also show a `curl` command, which you can run from the terminal (Erik Bern's key criterion for a demo). It's best for data/ML folks who want to be more full stack/self-serve, but also works to show that we're a prod deployment platform.

Screenshot 2025-10-30 at 5 53 03 PM

- `inference_perf.py` shows how to improve performance. It adds `Image.imports` + CPU Memory Snapshots and a Modal Volume to cache the weights. The Snapshots speed up  It also includes the `endpoint` functionality. It has to be`modal deploy`ed first, then run with `python` -- several times, manually stopping containers in between, if you want to show Memory Snapshot behavior/speedups. IME, the Memory Snapshots cut start times from 20-30s to under 10s. I skipped GPU Snapshots, since they are still in alpha.

Screenshot 2025-10-30 at 5 56 16 PM

## Bringing it all together: `inference_full.py`

The final example script brings all of the pieces together. I usually frame the `local_entrypoint` with the `map` as a test of the underlying Modal Function, since that's the pattern we're trying to demonstrate. The `main` block executed with Python is also intended for testing.

## What's still missing?

_There is no demonstration of Sandboxes_. The example suite is extensible to this though -- LLM-generated code is the killer app for Sandboxes. The cute thing to do would be to direct the LLM to edit the code and run it in a Sandbox, but I'm leaving that to future work.

There is no demonstration of Notebooks. I have some ideas here (call deployed Functions remotely from a Notebook) but I want to field test them more before making something shareable.

## Type of Change

## Monitoring Checklist

### DIFF
--- a/internal/typecheck.py
+++ b/internal/typecheck.py
@@ -63,7 +63,7 @@ def main() -> int:
                     )
                     future_to_path[future] = pth
 
-        for future in concurrent.futures.as_completed(future_to_path, timeout=60):
+        for future in concurrent.futures.as_completed(future_to_path, timeout=90):
             pth = future_to_path[future]
             try:
                 output = future.result()


charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-f388801.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-3ada065.modal.run

charlesfrye:
@prbot approve @shababo

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-385ec48.modal.run

shababo:
lgtm!

shababo:
>  I skipped GPU Snapshots, since they are still in alpha.

I think we can include this. Use of GPU snapshots, while still labelled as "experimental" are recommended to customers without much or any gating - in contrast to something like the distributed input plane.

You could add some prose explaining that GPU snapshots don't always work and to reach out if you have questions?

> The cute thing to do would be to direct the LLM to edit the code and run it in a Sandbox, but I'm leaving that to future work.

This is a good idea. I like the idea of starting with something and building out from it. I had thought the voice AI bot would also be good for this because it uses a lot of Modal patterns and features. But an LLM based application is definitely better.

PR #1386, opened 2025-11-04
remove min_containers where not necessary

charlesfrye:
Setting `min_containers` to reduce latency is common in production deployments but bad to include in examples, because it can be thoughtlessly copy-pasted, resulting in unexpected charges.

This PR drops `min_containers` from a number of examples where it was used, leaving it only in one example where it is required, due to the latency expectations of an external service. This example contains a note on costs. It also only uses a minimal CPU instance, so costs are contained (<$20/month, within the credits of the Starter Plan).

## Type of Change

### DIFF
--- a/01_getting_started/hello_world.py
+++ b/01_getting_started/hello_world.py
@@ -31,7 +31,7 @@ app = modal.App("example-hello-world")
 # [`@app.function`](https://modal.com/docs/reference/modal.App#function).
 
 
-@app.function(gpu="l40s")
+@app.function()
 def f(i):
     if i % 2 == 0:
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-81f248e.modal.run

charlesfrye:
🚀 The docs preview is ready! Check it out here:     https://modal-labs-examples--frontend-preview-9ea5823.modal.run

PR #1387, opened 2025-11-06
remove docs preview workflow

charlesfrye:
A recent change to our build system means that we no longer need to preview the frontend as part of updating the examples repo. In particular, deleting or renaming files is no longer as dangerous. Building a preview is still a polite thing to do when introducing a new example though!

This should cut typical CI times from multiple minutes (5-15) down to ~90s.

Modal engineers, see [this internal PR](https://github.com/modal-labs/modal/pull/29165) for more context.

## Type of Change


gleb-modal:
lgtm

PR #1388, opened 2025-11-06
reduce max inputs

charlesfrye:
This PR reduces the number of `max_inputs` on `modal.concurrent` decorators to 100 in places where it was 1000.

Writing a web server that can handle high load is hard, and we don't load-test our examples at that scale. We should stick with a smaller default, relying on auto-scaling above that. Users can opt in to higher per-replica scales if they wish, but the need to update the value provides an opportunity for them to consider what else might need to change.

## Type of Change


charlesfrye:
@prbot approve

mwaskom (approved):
This is 0.1x engineering at its finest.

PR #1391, opened 2025-11-10
Update huggingface-hub, speed up hf-xet

charlesfrye:
This PR updates the `huggingface-hub` package in our examples to be `0.36.0`, the last pre-1.0 release, apart from a few exceptions.

The `huggingface-hub` package is used in a lot of our examples and has been updated quite a bit over the last year or so.

Recent versions of the library use the `hf-xet` backend instead of `hf-transfer`. This PR updates our examples to use versions that use this backend. It also updates the environment variable settings (namely, `HF_XET_HIGH_PERFORMANCE` in place of `HF_HUB_ENABLE_HF_TRANSFER`) so that we continue to get maximum performance with the new backend.

## Type of Change


charlesfrye:
@prbot approve

charlesfrye:
@mwaskom a bunch of other package versions we use pin hf hub to <1.0, so bumping past it will take a bit more work. but this bump at least gets everyone on `hf_xet`.

PR #1392, opened 2025-11-11
you should be using uv

charlesfrye:
This PR migrates all of our examples to use `uv_pip_install` in place of `pip_install`.

Builds should be a bit faster now.

I turned off testing for one example, `dbt_duckdb`, which requires an AWS secret which was removed during an recent access audit. We should consider whether the example is worth the burden of maintaining that secret.

## Type of Change


charlesfrye:
Looks like something very cursed is broken in the `trtllm_latency` example. I'm going to force-merge this PR with that example broken and fix it in a follow-up.

PR #1393, opened 2025-11-11
revert change to trtllm_latency example install

charlesfrye:
Reverts the changed install method in #1392 until we get a chance to figure out why it broke.


charlesfrye:
@prbot approve

PR #1395, opened 2025-11-12
test dbt_duckdb

charlesfrye:
Now that we have the secret back, we can return to testing this example.

I also noticed a silent failure -- we changed the name of the bucket slightly, which leads the pipeline to fail silently. Fixed that as well.

## Type of Change


charlesfrye:
@prbot approve

parjohansson (approved):
lgtm

PR #1396, opened 2025-11-12
Make it compatible with external yaml files

turalaksel:
Execution of `modal run boltz_predict.py` with local yaml files leads to an error. Using `Path()` fixes the problem.


charlesfrye:
ooh, nice catch. thanks for the fix!

testing note: the Run changed examples Workflow fails due to missing perms on the fork. that's expected, so I ran it locally and it worked.

PR #1397, opened 2025-11-13
fix capitalization of Nsight

charlesfrye:
it really feels like the s should be capitalized in Nsight, but it is not.


charlesfrye:
@prbot approve

PR #1401, opened 2025-11-19
copy vllm_inference to llm_inference

charlesfrye:
Our flagship llm inference example may or may not use vllm, so let's rename it to llm_inference.

For now, just copying it so that we can do a separate redirection in the docs/marketing site, followed by deletion here.


charlesfrye:
@prbot approve

charlesfrye:
nvm, see #1402

PR #1402, opened 2025-11-19
remove extra llm_inference file

charlesfrye:
I came up with a better way to create an engine-agnostic `llm_inference` URL -- just add a layer of indirection with a redirect. No need to delete `vllm_inference`, so instead of doing that, I am reversing the changes in #1401.


charlesfrye:
@prbot approve

PR #1416, opened 2025-12-02
adds memory snapshotting to ministral3 example

charlesfrye:
This PR adds memory snapshotting to the vLLM Ministral 3 example.

Memory snapshots can improve startup times for compiled models running in vLLM by an order of magnitude - 3m to 20s in the case of Ministral 3 8B FP8, 2m to 12s in the case of Ministral 3 8B FP8.

They add a lot of complexity, though, as evinced by the code here. There are also some changes in behavior, e.g. the reduction in max seqs and seq lens in this code. So we don't really yet want to include snapshotting in our "baseline" vLLM example.

Note: I'm breaking the rules and loosely pinning vLLM here. I expect a release of vLLM shortly with version 0.11.3 and improved support for Ministral 3. A loose pin lets users who run the example after that release get those benefits. I'm also not pinning the model revision, because I similarly expect release imminently.

### DIFF
--- a/06_gpu_and_ml/llm-serving/ministral3_inference.py
+++ b/06_gpu_and_ml/llm-serving/ministral3_inference.py
@@ -1,10 +1,19 @@
-# # Serve Ministral 3 with vLLM
+# ---
+# deploy: true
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve

PR #1417, opened 2025-12-02
remove hf secret from ministral 3 example

charlesfrye:
The weights are now public and ungated, so the Secret is no longer needed


shababo:
lgtm!

PR #1428, opened 2025-12-29
simple code interpreter fix

charlesfrye:
Rejigger the `simple_code_interpreter` example to use a single Iterator over the `StreamReader` of the `ContainerProcess.stdout` for the interpreter's driver program. This avoids the `StopIteration` we currently get for this code on recent versions of the Modal client.

In short, what was

```python
def run_code(p: modal.container_process.ContainerProcess, code: str):
```

becomes

```python
def run_code(writer: modal.io_streams.StreamWriter, reader: Iterator[str], code: str):
```

### DIFF
--- a/13_sandboxes/simple_code_interpreter.py
+++ b/13_sandboxes/simple_code_interpreter.py
@@ -20,6 +20,7 @@
 
 import inspect
 import json
+import sys
 from typing import Any, Iterator
 
 import modal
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve @saltzm

saltzm:
LGTM. @freider did we ever figure out why StopIteration was being returned?

PR #1429, opened 2026-01-02
update sglang low latency

charlesfrye:
This PR updates the SGLang low latency example, which uses the new `experimental.http_server`.

Some notes:
- complete explanatory text, with a focus on the knobs in SGLang and Modal that can reduce latency.
- no more snapshotting. I couldn't get snapshots to work with multiple GPUs. snapshotting is not as critical for the use cases targeted by this example, which we presume have a high enough load that they don't need to scale from 0.
- FP8 version of Qwen 3, for reduced latency.
- `H100!:2` -- H100s are more plentiful than B200s and so pinning to a specific region doesn't lead to queues. we want more than one GPU so that we can go tensor parallel (30% improvement over baseline)
- the latest version of SGLang uses JIT-compiled DeepGEMM kernels for FP8 on Hopper, so we need to handle DeepGEMM's cache on a Volume.

Screenshot 2026-01-02 at 4 34 29 PM

### DIFF
--- a/06_gpu_and_ml/llm-serving/sglang_low_latency.py
+++ b/06_gpu_and_ml/llm-serving/sglang_low_latency.py
@@ -1,9 +1,4 @@
-# ---
-# deploy: true
-# cmd: ["python", "06_gpu_and_ml/llm-serving/sglang_low_latency.py"]
-# ---
-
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve

PR #1430, opened 2026-01-03
add specdec to sglang_low_latency

charlesfrye:
Adds speculative decoding with [Tengyunw's EAGLE-3 model](https://huggingface.co/Tengyunw/qwen3_8b_eagle3) to our low latency SGLang example.

This does slow down server startup quite a bit, from ~2m to ~3m -- the warmup calls now take about a minute instead of a few seconds. I didn't dive in.

Unfortunately we need the tensor-parallel config to do the AoT compilation of DeepGEMM kernels, which makes the order of explanation and code a bit jank -- we discuss GPU config at the top but don't talk about why until later. But I think it's better to do it all together.

Screenshot 2026-01-03 at 3 38 51 PM

### DIFF
--- a/06_gpu_and_ml/llm-serving/sglang_low_latency.py
+++ b/06_gpu_and_ml/llm-serving/sglang_low_latency.py
@@ -39,28 +39,17 @@ sglang_image = (
     ).entrypoint([])  # silence chatty logs on container start
 )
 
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve @molocule

PR #1432, opened 2026-01-05
update vllm_inference - version and model

charlesfrye:
We like to keep this one fresh! I wasn't able to find a more recent 4-8B FP8 model with better reviews than Qwen 3 2507, so let's stick with that.

### DIFF
--- a/06_gpu_and_ml/llm-serving/vllm_inference.py
+++ b/06_gpu_and_ml/llm-serving/vllm_inference.py
@@ -2,7 +2,7 @@
 # pytest: false
 # ---
 
-# # Run OpenAI-compatible LLM inference with Qwen3-8B and vLLM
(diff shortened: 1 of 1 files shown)


charlesfrye:
Note: vLLM boots are slower now -- 60s to 80s, when we'd previously gotten it down to ~45s [back when we were using version `0.9`](https://github.com/modal-labs/modal-examples/pull/1217/changes). I think that's due to changes in vLLM across versions, but haven't root-caused.

charlesfrye:
The new model really loves to think -- the "Jar Jar Binks explains the SVD" causes it to generate a few thousand tokens of hemming and hawing. That makes the runtime of the test noticeably longer.

charlesfrye:
@prbot approve

PR #1439, opened 2026-01-13
log decode tok/s more frequently

charlesfrye:
At the lower logging rate, we under-estimate peak tok/s by a lot. But the default logging rate is too high, so we set it in between.

## Type of Change


charlesfrye:
@prbot approve

PR #1405, opened 2025-11-23
Bump vLLM examples to 0.14.0 and unify CUDA/Python base images

mmangkad:
- bump every vllm example to latest vllm versions
- standardize all vllm-serving images on cuda 12.8.1 registry bases
- tune the gpt-oss example for blackwell b200 gpus with the recommended optimized settings
- migrate the cuda graph configuration to the new `--max-cudagraph-capture-size` api

PR #1442, opened 2026-01-15
better config, update vllm, >250 tok/s

charlesfrye:
gpt-oss go brrrt

Adds a speculator I found online and a few handy flags from the [vLLM Recipes](https://docs.vllm.ai/projects/recipes/en/latest/OpenAI/GPT-OSS.html) docs.

Thanks @mmangkad for clueing me into the need to store FlashInfer JIT kernels with the new vLLM version, that was killing start times. And also for the nudge to upgrade via PR #1405


charlesfrye:
@prbot approve @emilyhanyf

emilyhanyf (commented):
👍

PR #1445, opened 2026-01-20
adds Modal-Session-ID header and explanation

charlesfrye:
Updates the `sglang_low_latency` example to include the `Modal-Session-ID` header in the client, along with an explanation of why it's there and how to use it.

Also two minor cleanups: formatting and a comment on model revision pinning.

## Type of Change

### DIFF
--- a/06_gpu_and_ml/llm-serving/sglang_low_latency.py
+++ b/06_gpu_and_ml/llm-serving/sglang_low_latency.py
@@ -61,8 +61,8 @@ GPU = f"{GPU_TYPE}:{N_GPUS}"
 # in each forward pass.
 
 MODEL_NAME = "Qwen/Qwen3-8B-FP8"
-MODEL_REVISION = (
-    "220b46e3b2180893580a4454f21f22d3ebb187d3"  # latest commit as of 2026-01-01
+MODEL_REVISION = (  # pin revision id to avoid nasty surprises!
+    "220b46e3b2180893580a4454f21f22d3ebb187d3"  # latest commit as of 2026-01-20, from 2025-07-25
 )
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve

06_gpu_and_ml/llm-serving/sglang_low_latency.py
```diff
@@ -479,13 +486,30 @@ async def test(test_timeout=10 * MINUTES, prompt=None, twice=True):
...
+# The value associated with this key
+# is used to map requests onto containers such that
+# while the set of containers is fixed, requests with the same value
+# are sent to the same container.
+# Set this to a different value per multi-turn interaction
+# (prototypically, a user conversation thread with a chatbot)
+# to improve KV cache hit rates.
+# Additionally, when the set of containers changes (e.g. due to autoscaling),
+# sessions are rebalanced such that load is approximately evenly spread,
+# much like in [RAID rebalancing](https://cordero.me/understanding-raid-rebalance-ensuring-optimal-performance-and-data-protection/).
```
molocule, line 500: should we just say via rendezvous hashing since it's a canonical proxy setup?

charlesfrye: I'm loath to commit to a specific algorithm or technique, because we might (and should reserve the right to!) change it. Using an analogy to another domain's solution to the same problem decouples the explanation from our implementation.

I think RAID is sufficiently well-known that it's not distracting here. And maybe it's just my background, but I think disk rebalancing is a simpler/more obvious case than proxy load-balancing.

PR #1451, opened 2026-01-26
put a stricter limit in place for document length

charlesfrye:
A new document in the SEC EDGAR Feed led to this example failing. We don't explicitly limit the token count in documents, just the character count, and our conservative heuristic failed, resulting in a prompt with token count 17,149, above the `max_model_len` of 16,384.

If we limit the documents to approximately 15k characters, they can only produce a document with token count at most a bit over 15k. That leaves room for 1k tokens of output and the short prompt in our `max_model_len`.


charlesfrye:
@prbot approve

shababo:
lgtm

PR #1456, opened 2026-01-27
skip links that are slow to load

charlesfrye:
Adds a short timeout and skip logic for the playwright page load.

Noticed some flakes in this example, and they appear to come from pages taking a long time to load.

We don't have control of the loaded pages -- and you generally don't when running a webscraper! -- so we need to skip here.


charlesfrye:
@prbot approve

shababo:
lgtm

PR #1457, opened 2026-01-28
Fix WebRTC YOLO: make standalone and TURN optional

aaronsuns:
- Change relative imports to absolute imports for standalone usage. When Modal CLI parses the file locally, relative imports fail with `ImportError: attempted relative import with no known parent package`. Absolute imports work both when run as a module and when parsed standalone.

- Make TURN credentials optional (works with free STUN servers). The original code required TURN credentials to deploy, which creates unnecessary friction. Most users can use free STUN servers.
  - Removed required secret from `@app.cls` decorator
  - Added graceful fallback in `get_turn_servers()` method
  - Added Google STUN server as backup option
  - Only adds TURN servers if credentials are available in environment

This allows the example to work out-of-the-box without requiring users to set up TURN credentials first.


07_web_endpoints/webrtc/webrtc_yolo.py
```diff
@@ -110,7 +110,13 @@
...
 import modal
 
-from .modal_webrtc import ModalWebRtcPeer, ModalWebRtcSignalingServer
+# Handle imports for both module execution (-m) and standalone parsing
+try:
+    # Try relative imports first (works when run as module with -m)
+    from .modal_webrtc import ModalWebRtcPeer, ModalWebRtcSignalingServer
+except ImportError:
+    # Fall back to absolute imports (works when Modal CLI parses standalone)
+    from modal_webrtc import ModalWebRtcPeer, ModalWebRtcSignalingServer
```
shababo, line 119: You have to deploy with `-m` or else the additional files won't be available on the container and this import will fail then.

07_web_endpoints/webrtc/webrtc_yolo.py
```diff
@@ -194,11 +200,17 @@
...
+# For now, we'll use an empty secrets list - TURN servers will be skipped if credentials aren't available
+# You can add the secret later: modal secret create turn-credentials TURN_USERNAME=... TURN_CREDENTIAL=...
+
+
 @app.cls(
     image=video_processing_image,
     gpu="A100-40GB",
     volumes=cache,
-    secrets=[modal.Secret.from_name("turn-credentials")],
+    # secrets=[modal.Secret.from_name("turn-credentials")],  # Uncomment when you create the secret
```
shababo, line 213: Can you add a `try` statement that attempts instantiate `turn_creds = [modal.Secret.from_name("turn-credentials")]` and if it errors we set `turn_creds = None`? Then we don't need to have this optional commenting mechanism.

07_web_endpoints/webrtc/webrtc_yolo.py
```diff
@@ -320,7 +341,11 @@ async def root():
 def get_yolo_model(cache_path):
     import onnxruntime
 
-    from .yolo import YOLOv10
+    # Handle imports for both module execution (-m) and standalone parsing
+    try:
+        from .yolo import YOLOv10
+    except ImportError:
+        from yolo import YOLOv10
```
shababo, line 348: Same as above... we have to use the relative import and `-m`.

07_web_endpoints/webrtc/webrtc_yolo.py
```diff
@@ -337,7 +362,11 @@ def get_yolo_track(track, yolo_model=None):
     from aiortc import MediaStreamTrack
     from aiortc.contrib.media import VideoFrame
 
-    from .yolo import YOLOv10
+    # Handle imports for both module execution (-m) and standalone parsing
+    try:
+        from .yolo import YOLOv10
+    except ImportError:
+        from yolo import YOLOv10
```
shababo, line 369: Same.

shababo (changes requested):

PR #1461, opened 2026-01-29
allow HF Hub access when using dummy weights

charlesfrye:
Quick followup to #1460 -- realized that if you haven't previously downloaded the whole repo with the full weights, you do need to reach the HF_HUB to run the dummy weights (to fetch config).


charlesfrye:
@prbot approve

jvmncs:
lgtm

PR #1466, opened 2026-02-01
Fix ImportError: libxcb.so.1

softwarebouwer:
Fix ImportError: libxcb.so.1: cannot open shared object file: No such file or directory by installing libxcb1, libxcb-xinerama0, libxcb-cursor

## Problem
The container failed to start 
```
ImportError: libxcb.so.1: cannot open shared object file: No such file or directory by installing libxcb1, libxcb-xinerama0, libxcb-cursor
```

## Root Cause
vLLM commit ee2129182 (Jan 5, 2026) added Nemotron Parse model with top-level import cv2 in vllm/model_executor/models/nemotron_parse.py:14. This causes cv2 to be imported for ALL models during startup, not just Nemotron Parse. The cv2 import requires X11 client libraries (libxcb.so.1) which are missing from the Docker image.

source https://github.com/LMCache/LMCache/issues/2452

## Solution
install libxcb1, libxcb-xinerama0, libxcb-cursor

## Changes
Added `.apt_install("libxcb1", "libxcb-xinerama0", "libxcb-cursor0")`
Pin `vllm==0.15.0`
Pin `flashinfer-python==0.6.1`

## Type of Change

## Monitoring Checklist

- [x] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`) `modal deploy ministral3_inference.py`

### Content

### Build Stability

## Outside Contributors

### DIFF
--- a/06_gpu_and_ml/llm-serving/ministral3_inference.py
+++ b/06_gpu_and_ml/llm-serving/ministral3_inference.py
@@ -38,10 +38,11 @@ app = modal.App("example-ministral3-inference")
 vllm_image = (
     modal.Image.from_registry("nvidia/cuda:12.9.0-devel-ubuntu22.04", add_python="3.12")
     .entrypoint([])
+    .apt_install("libxcb1", "libxcb-xinerama0", "libxcb-cursor0")
     .uv_pip_install(
-        "vllm==0.13.0",
+        "vllm==0.15.0",
         "huggingface-hub==0.36.0",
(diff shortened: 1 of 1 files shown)


charlesfrye:
Thanks for sharing!

As indicated in your RCA, this only impacts vLLM versions later than the one currently used in the example. I am hoping this dependency will be removed in a future version.

PR #1467, opened 2026-02-02
better test prompt

charlesfrye:
This test prompt produces a lot more output (evincing decode perf improvements) and also triggers a variety of specdec accept lengths.


charlesfrye:
@prbot approve

PR #1468, opened 2026-02-02
add longer timeouts, fix edgartools pytz issue

charlesfrye:
The `vllm_throughput` example was failing in synmons. The pipeline is not as robust to failures as it could be.

One common failure is from timeouts, resolved in this PR. As written, I actually think this example would fail on first run for users in most cases, since the `orchestrate` Function usually takes more than five minutes if the data isn't already in the Volume.

I'm leaving off fixing the underlying reliability problem for now, since that's less user-facing. Will keep an eye on synthetic monitoring and hope to catch more failure modes.

Separately, the example was failing during rebuild ([sample failure](https://github.com/modal-labs/modal-examples/actions/runs/21503073888/job/61953393313)). The cause was a transitive dependency of a dependency -- [pandas 3 dropped pytz and edgartools uses pytz](https://www.edgartools.io/pandas-3-0-and-edgartools/). This PR bumps our edgartools version to fix this.

### DIFF
--- a/06_gpu_and_ml/llm-serving/vllm_throughput.py
+++ b/06_gpu_and_ml/llm-serving/vllm_throughput.py
@@ -348,7 +348,13 @@ class Vllm:
 # in our project by putting it in a separate container Image.
 
 data_proc_image = modal.Image.debian_slim(python_version="3.13").uv_pip_install(
-    "edgartools==5.11.1"
+    # pin transitive deps to avoid surprises like this one:
+    # https://www.edgartools.io/pandas-3-0-and-edgartools/
+    "edgartools==5.8.3",
+    "httpx==0.28.1",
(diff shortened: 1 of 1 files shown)


charlesfrye:
There was _another_ transitive dep issue, revealed by fixing the existing issue.

```python
TypeError: Limiter.__init__() got an unexpected keyword argument 'max_delay'
```

This time it was for `pyrate-limiter` as used by `httpxthrottlecache` as used by `httpx` as used by `edgartools`.

Solved it in the end by `modal shell`ing into a fresh container with the broken deps and examining the `site-packages` to find the offending import, then checking a past successful run for the last working version of `pyrate-limiter`.

charlesfrye:
@prbot approve

PR #1469, opened 2026-02-03
skip overwriting config yaml

charlesfrye:
We were _always_ overwriting the default config YAML file for the `very_large_models` example. This PR stops doing that.

It's bad for users, who probably expect to be able to change things in that file and have those changes respected, and it's bad for the synmon environment, which has a read-only filesystem outside of `/tmp`.


charlesfrye:
@prbot approve

PR #1470, opened 2026-02-03
remove llama cpp example

charlesfrye:
This PR removes the `llama_cpp` example.

Most users will deploy with vLLM or SGLang, not llama.cpp. This example was created in part because llama.cpp had better early support for ternary DeepSeek, which was important before we added H200s and B200s to the platform and so could more easily serve higher precision versions.

The `very_large_models` example replaces it for the purpose of demonstrating how to run large models on Modal.


charlesfrye:
@prbot approve

jvmncs:
lgtm

PR #1472, opened 2026-02-04
adds a demo of opencode running headlessly in a Modal Sandbox

charlesfrye:
This PR adds a basic demonstration of OpenCode in a Modal Sandbox.

We run the headless `opencode serve` command and then use the blessed `opencode attach` TUI and browser interfaces.

To show patterns for including more permissions and code, I pull in the examples. Which means this example can run itself!

examples-opencode-server-webui


charlesfrye:
@prbot approve

PR #1473, opened 2026-02-04
add prose to vllm_low_latency, match sglang_low_latency

charlesfrye:
This PR ports the prose and other adaptations added to the `sglang_low_latency` example after its initial merge over to the `vllm_low_latency` example.

Co-authored by Claude Opus 4.5.

### DIFF
--- a/06_gpu_and_ml/llm-serving/vllm_low_latency.py
+++ b/06_gpu_and_ml/llm-serving/vllm_low_latency.py
@@ -255,6 +255,7 @@ app = modal.App(name=APP_NAME)
     experimental_options={"enable_gpu_snapshot": True},
     region=REGION,
     min_containers=MIN_CONTAINERS,
+    timeout=10 * MINUTES,
 )
 @modal.experimental.http_server(
     port=PORT,  # wrapped code must listen on this port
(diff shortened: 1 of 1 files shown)


06_gpu_and_ml/llm-serving/vllm_low_latency.py
```diff
@@ -3,84 +3,270 @@
...
+            requests.get(f"http://127.0.0.1:{PORT}/health").raise_for_status()
+            return
+        except (
+            subprocess.CalledProcessError,
+            requests.exceptions.ConnectionError,
+            requests.exceptions.HTTPError,
+        ):
+            time.sleep(5)
+    raise TimeoutError(f"vLLM server not ready within {timeout} seconds")
+
```
molocule, line 216: we technically don't need this anymore, but still good to have

06_gpu_and_ml/llm-serving/vllm_low_latency.py
```diff
@@ -3,84 +3,270 @@
...
     image=vllm_image,
-    gpu="H100",
+    gpu=GPU,
     volumes={HF_CACHE_PATH: HF_CACHE_VOL},
     enable_memory_snapshot=True,
     experimental_options={"enable_gpu_snapshot": True},
-    region="us-east",
+    region=REGION,
     min_containers=MIN_CONTAINERS,
-    timeout=6 * 60,
```
molocule, line 70: don't we still need this?

06_gpu_and_ml/llm-serving/vllm_low_latency.py
charlesfrye: wait, why not? don't we need this before we call `warmup` as part of snapshotting?

PR #1542, opened 2026-04-15
remove GPU packing example

charlesfrye:
Fresh builds for this example broke recently. I tried having an agent fix, but it got confused and spewed a bunch of noise.

Stepping back: this example was always a bit dubious. We sleep to mock slow work. But if you actually do work, you introduce contention for resources between the replicas. We have tried a few times to find a case where real work actually enjoys a benefit from packing, and we haven't found it. You need really big host bubbles and/or kernels that don't fill all the SMs to even have a chance, and we just don't see these cases in practice.

So rather than fix the narrow-sense bug in builds here, I am just going to delete until we find a case where GPU packing makes sense. cc @olokobayusuf

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1542" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


charlesfrye:
@prbot approve

charlesfrye:
@prbot approve @shababo 

(sorry, dunteman isn't in the prbot system yet)

shababo:
lgtm!

100% agree. Unfortunately `asyncio.sleep` does not properly replicate the async nature of CUDA calls - which is not necessarily free or even easy to take advantage of. It overstates the simplicity and gains :(

PR #1568, opened 2026-05-20
move Python pkg install into one step for ltx2 two stage

charlesfrye:
Just because the model is two stage doesn't mean our install should be!

This examples builds were failing in synmon. Root cause was version divergence from using `uv pip` twice. Versions are not preserved across invocations.

So this PR moves it all into a single command, which requires `unsafe-best-match` to pull packages from both our mirror and the Pytorch index.

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1568" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/06_gpu_and_ml/text-to-video/ltx2_two_stage.py
+++ b/06_gpu_and_ml/text-to-video/ltx2_two_stage.py
@@ -38,7 +38,9 @@ from pathlib import Path
 import modal
 
 # We pin an LTX-2 commit and install its three subpackages.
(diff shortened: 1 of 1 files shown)


charlesfrye:
@prbot approve

BenCowen:
Thanks, I didn't know this issue could arise. I will keep this in mind for future demos as well!

BenCowen (commented):
looks good

PR #1574, opened 2026-05-28
internal/deploy: fix incorrect cwd when deploying example modules

Dev-X25874:
`deploy.py` passed `Path(ex_mod.module)` to the `deploy()` function, where
`ex_mod.module` is a dotted Python import path like `06_gpu_and_ml.text_to_speech`.
Wrapping that in `Path(...)` means `.parent` resolves to `.` (the current working
directory) instead of the actual example subdirectory. As a result, `subprocess.run`
was executing `modal deploy` from the wrong directory, which would cause Modal to
fail to resolve the module.

**Fix:** pass `Path(ex_mod.filename)` instead, so `.parent` correctly points to the
example's subdirectory, and use `.stem` for the `-m` flag so the extension is stripped.

Added `internal/test_deploy.py` with three tests that cover the corrected `cwd`,
the correct module flag, and the no-op path for non-deployable examples.

## Type of Change

## Monitoring Checklist

N/A — this change is to `internal/deploy.py`, not to any example.

## Documentation Site Checklist

N/A — no example files were added or modified.

## Outside Contributors

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1574" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/internal/deploy.py
+++ b/internal/deploy.py
@@ -82,7 +82,7 @@ def main(argv: Optional[list[str]] = None) -> int:
     results = [
         deploy(
             deployable=bool(ex_mod.metadata.get("deploy")),
-            module_with_app=Path(ex_mod.module),
+            module_with_app=Path(ex_mod.filename),
             dry_run=arguments.dry_run,
             filter_pttrn=filter_pttrn,
             env=ex_mod.metadata.get("env"),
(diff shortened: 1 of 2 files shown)


charlesfrye:
I don't see any evidence that the Deploy workflow is failing, so I don't think this is a real bug.

Dev-X25874:
Hey @charlesfrye — sorry about the noise on this one. I should have verified the Deploy workflow was actually failing before opening the PR. Thanks for taking the time to look into it and for the clear feedback. I'll make sure to do more thorough investigation next time before submitting a fix.

PR #1583, opened 2026-06-03
Text-to-world video generation with LTX-2.3 and InSpatio

zhouhelena1:
This example chains two diffusion models on Modal to turn a text prompt into an explorable 3D "world": a video rendered along a camera trajectory, plus the point cloud it was built from. Here's an [example of a generated video](https://modal-cdn.com/world.mp4).

image

The pipeline runs in two stages:
1. [LTX-2.3](https://huggingface.co/Lightricks/LTX-2.3) generates a short reference video from the prompt/image.
2. [InSpatio-World](https://huggingface.co/inspatio/world) lifts that video into a 3D scene and re-renders it along a specified camera trajectory.

## Type of Change

- [X] New example for the GitHub repo
  - [X] New example for the documentation site (Linked from a discoverable page, e.g. via the sidebar in `/docs/examples`)

## Monitoring Checklist

    - [X] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`)
    - [X] Example is tested by running the `cmd` with no arguments, or the `args` are provided in the example frontmatter (e.g. `args: ["--prompt", "Formula for room temperature superconductor:"]`
    - [X] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or other code executed locally)

### Content
  - [X] Example is documented with comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style
  - [X] All media assets for the example that are rendered in the documentation site page are retrieved from `modal-cdn.com`

### Build Stability
  - [X] Example pins all dependencies in container images
    - [X] Example pins container images to a stable tag like `v1`, not a dynamic tag like `latest`
    - [X] Example specifies a `python_version` for the base image, if it is used 
    - [X] Example pins all dependencies to at least [SemVer](https://semver.org/) minor version, `~=x.y.z` or `==x.y`, or we expect this example to work across major versions of the dependency and are committed to maintenance across those versions
      - [X] Example dependencies with `version < 1` are pinned to patch version, `==0.y.z`

## Outside Contributors


06_gpu_and_ml/text-to-world/image_to_world.py
BenCowen: My guess is that these constants can't actually be fiddled with. So I'd recommend simplifying and just hard-coding the names in the enter method.

06_gpu_and_ml/text-to-world/image_to_world.py
```diff
@@ -445,32 +448,32 @@ def git_lfs_clone(url: str, dest: str):
...
 
-        git_lfs_clone("https://huggingface.co/inspatio/world", "world")
+        def git_clone(url: str, dest: str):
+            dest_path = cwd / dest
+            if dest_path.exists():
+                shutil.rmtree(dest_path)
+            subprocess.run(["git", "clone", url, dest], cwd=cwd, env=env, check=True)
+
+        # Download repos concurrently on first container start
+        clone_tasks = [
```
BenCowen: is this the simplest way to do this? I would expect you could use the same huggingface download api that you used for LTX.

using the huggingface api is ideal because it will manage the model repos for you so you wouldn't need _link_checkpoints etc

06_gpu_and_ml/text-to-world/image_to_world.py
BenCowen: I would just use one volume for all model caching

BenCowen: pro tip: you can mount the huggingface cache directly to `/root/.cache/huggingface` which removes the constant / need to manage the path by variable name. See here:
https://github.com/modal-labs/modal-examples/blob/b9b9d67f5c54261746b9d435ef0c29b751295eee/06_gpu_and_ml/text-to-video/ltx2_two_stage.py#L97

BenCowen, line 342: I think it's safer to pull the torch and flash-attn installs into this `uv_pip_install`, as you did for `ltx_image`

06_gpu_and_ml/text-to-world/image_to_world.py
```diff
@@ -428,6 +430,7 @@ class InSpatioInference:
     def setup(self):
         """Download InSpatio checkpoints to the weights Volume on first container start."""
         import os
+        from concurrent.futures import ThreadPoolExecutor
```
BenCowen: convention is to import python basics at the top of the file rather than in the specific method

06_gpu_and_ml/text-to-world/image_to_world.py
BenCowen: I would like to walk through this part with you when you have a few minutes. In summary, I'd recommend you don't load or pass any bytes around like this. Just (1) save the output to a volume in the ltx function, and (2) pass Inspatio the volume path where the video was saved (ie just a string); (3) the inspatio function can just load the mp4 (or whatever) directly from the volume. 

Hopefully this allows you to remove a bunch of the manifest and bytes conditioning stuff.

06_gpu_and_ml/text-to-world/image_to_world.py
```diff
@@ -0,0 +1,1005 @@
+# ---
+# output-directory: "/tmp/world_model"
+# args: ["--prompt", "A serene mountain lake at sunrise, mist rising off the water"]
+# ---
+
+# # Image-to-world video generation with LTX-2.3 and InSpatio
+
+# This example chains two diffusion models on Modal to turn a text prompt
+# (and an optional reference image) into an explorable 3D "world": a video
+# rendered along a camera trajectory, plus the point cloud it was built from.
```
charlesfrye: Could we get a screenshot or video of the interaction with the world? Upload the file to S3 `modal-public-assets` aka `modal-cdn.com`, cf some of the other examples.

BenCowen:
my bad- i accidentally clicked ready for review but we still need to add the world asset

charlesfrye:
Had some minor changes that felt easier to just express as a PR, #1585.

Once that's merged here, happy to merge the result into `main`.

charlesfrye (approved):
LGTM!

I don't love that there's no way to interact with the "world" here -- it's just a video, right? But let's merge regardless (this is a common issue with world model demos) and come back if there's more interest.

BenCowen:
I was also thinking it would be really nice if the user could input the trajectory, or click and drag etc...

PR #1580, opened 2026-06-02
Update examples to use Modal Server

molocule:
## Type of Change

- [ ] New example for the GitHub repo
  - [ ] New example for the documentation site (Linked from a discoverable page, e.g. via the sidebar in `/docs/examples`)
- [ ] Example updates (Bug fixes, new features, etc.)
- [ ] Other (Changes to the codebase, but not to examples)

## Monitoring Checklist

  - [ ] Example is configured for testing in the synthetic monitoring system, or `lambda-test: false` is provided in the example frontmatter and I have gotten approval from a maintainer
    - [ ] Example is tested by executing with `modal run`, or an alternative `cmd` is provided in the example frontmatter (e.g. `cmd: ["modal", "serve"]`)
    - [ ] Example is tested by running the `cmd` with no arguments, or the `args` are provided in the example frontmatter (e.g. `args: ["--prompt", "Formula for room temperature superconductor:"]`
    - [ ] Example does _not_ require third-party dependencies besides `fastapi` to be installed locally (e.g. does not import `requests` or `torch` in the global scope or other code executed locally)

### Content
  - [ ] Example is documented with comments throughout, in a [_Literate Programming_](https://en.wikipedia.org/wiki/Literate_programming) style
  - [ ] All media assets for the example that are rendered in the documentation site page are retrieved from `modal-cdn.com`

### Build Stability
  - [ ] Example pins all dependencies in container images
    - [ ] Example pins container images to a stable tag like `v1`, not a dynamic tag like `latest`
    - [ ] Example specifies a `python_version` for the base image, if it is used 
    - [ ] Example pins all dependencies to at least [SemVer](https://semver.org/) minor version, `~=x.y.z` or `==x.y`, or we expect this example to work across major versions of the dependency and are committed to maintenance across those versions
      - [ ] Example dependencies with `version < 1` are pinned to patch version, `==0.y.z`

## Outside Contributors

You're great! Thanks for your contribution.

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1580" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #1594, opened 2026-06-22
Add unauthenticated=True to nemotron http_server to fix 401 in CI

roscioli:
## Summary

- Adds `unauthenticated=True` to the `@modal.experimental.http_server` decorator in the nemotron example so the service is publicly accessible without a `Modal-Authorization` header.

modal-labs/modal@0a7389485d changed `http_server` services (those with `is_server=True`) to require proxy auth by default when `http_config.unauthenticated=False`. The nemotron example is a public demo and the test code makes unauthenticated requests, which started returning 401 after that change.

This also requires modal-labs/modal#XXXXX to expose `unauthenticated` as a parameter on `modal.experimental.http_server()` — the two changes should land together.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1594" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/06_gpu_and_ml/llm-serving/nemotron_inference.py
+++ b/06_gpu_and_ml/llm-serving/nemotron_inference.py
@@ -236,6 +236,7 @@ PORT = 8000
     port=PORT,  # wrapped code must listen on this port
     proxy_regions=[PROXY_REGION],  # location of proxies, should overlap with Cls region
     exit_grace_period=15,  # seconds, time to finish up requests when closing down
+    unauthenticated=True,  # allow public access without a Modal-Authorization token
 )
 @modal.concurrent(target_inputs=TARGET_INPUTS)
 class Server:


06_gpu_and_ml/llm-serving/nemotron_inference.py
```diff
@@ -292,47 +229,45 @@ def warmup():
     volumes={HF_CACHE_PATH: HF_CACHE_VOL},
     region=REGION,
     min_containers=MIN_CONTAINERS,
-    secrets=[hf_secret],
-    startup_timeout=120 * MINUTES,  # time to load weights
+    secrets=[modal.Secret.from_name("huggingface-secret")],
+    startup_timeout=20 * MINUTES,  # time to load weights
 )
 @modal.experimental.http_server(
```
thomasjpfan, line 235: Can we migrate this to use `@app.server`?

charlesfrye:
I believe this is subsumed by [this PR](https://github.com/modal-labs/modal-examples/pull/1580).

PR #1595, opened 2026-06-22
add basic harbor sandbox eval example

andrewhinh:
Add a new starter Sandboxes example to demonstrate how Sandboxes work in the SDK and dashboard.


aaazzam (commented):
This is a bit involved for a Sandbox 101. Most other getting_started examples are roughly 20-30 lines of code and sprint towards showing the core primitive. I think this example could be repurposed for something more eval oriented (running evals with harbor or something) but as-is I think this is hard to _lead off with_. What do you think?

andrewhinh:
> This is a bit involved for a Sandbox 101. Most other getting_started examples are roughly 20-30 lines of code and sprint towards showing the core primitive. I think this example could be repurposed for something more eval oriented (running evals with harbor or something) but as-is I think this is hard to _lead off with_. What do you think?

this was bascially a first draft response to this:
> Tactically: I like to show the SDK and dashboard via a demo, `inference.py` and `inference_map.py` or `get_started.py` with H100s and `map` added live, scaling up to 50 or so GPUs. Showing massive scaling tends to impress people and get the vague shape of the value prop in their heads. You can see a recorded version of the `get_started` demo from January 26, 2025 [here](https://youtu.be/Y7n8sIZV1vQ?t=26). The SDK and dashboard are our primary product surfaces, so you want to use this opportunity to familiarize people with them — even if just vaguely, so that the next impression sticks better. [We Should™️](https://app.notion.com/p/Modal-Sabotage-Field-Manual-2001e7f1694980d3bf90e276643b5c9d?pvs=21)  come up with something similar for Sandboxes — probably scaling up to 100 or 1000 containers running a coding agent, then having an LLM judge pick the best outcome? I have an old demo in this direction [here](https://github.com/modal-labs/mistral-human-eval/), but it’s too complicated to use in most settings.

agree tho it's too long, especially all the test setup. but actually I didn't know about harbor, let me play around with it.

aaazzam:
One concrete thing worth flagging on top of the scope discussion: this defaults to n_parallel=1000, but Modal's Starter plan caps out at 100 concurrent containers, and Team tops out at 1000. Most people trying this on a Starter account queue behind their account's cap instead of getting the 1000-way parallelism it's showing off. There's also no frontmatter overriding args here, so the repo's own scheduled test runner launches this with that same default of 1000 on every run once it merges, the same way a curious reader's first try would.

Worth dialing the default down to something small enough to run cleanly on any tier, maybe single digits, and checking what concurrency ceiling fits whatever tier this example is meant to target before picking the number meant to impress.


> > This is a bit involved for a Sandbox 101. Most other getting_started examples are roughly 20-30 lines of code and sprint towards showing the core primitive. I think this example could be repurposed for something more eval oriented (running evals with harbor or something) but as-is I think this is hard to _lead off with_. What do you think?
> 
> this was bascially a first draft response to this:
> 
> > Tactically: I like to show the SDK and dashboard via a demo, `inference.py` and `inference_map.py` or `get_started.py` with H100s and `map` added live, scaling up to 50 or so GPUs. Showing massive scaling tends to impress people and get the vague shape of the value prop in their heads. You can see a recorded version of the `get_started` demo from January 26, 2025 [here](https://youtu.be/Y7n8sIZV1vQ?t=26). The SDK and dashboard are our primary product surfaces, so you want to use this opportunity to familiarize people with them — even if just vaguely, so that the next impression sticks better. [We Should™️](https://app.notion.com/p/Modal-Sabotage-Field-Manual-2001e7f1694980d3bf90e276643b5c9d?pvs=21)  come up with something similar for Sandboxes — probably scaling up to 100 or 1000 containers running a coding agent, then having an LLM judge pick the best outcome? I have an old demo in this direction [here](https://github.com/modal-labs/mistral-human-eval/), but it’s too complicated to use in most settings.
> 
> agree tho it's too long, especially all the test setup. but actually I didn't know about harbor, let me play around with it.

One concrete thing worth flagging on top of the scope discussion: this defaults to n_parallel=1000, but our Starter plan caps out at 100 concurrent containers, and Team tops out at 1000. Most people trying this on a Starter account queue behind their account's cap instead of getting the 1000-way parallelism it's showing off. There's also no frontmatter overriding args here, so the repo's own scheduled test runner will burn a few hundred sandboxes unnecessarily.

Worth dialing the default down to something small enough to run cleanly on any tier?

uv.lock
charlesfrye, line 1: This file should be removed.

13_sandboxes/harbor_evals.py
```diff
@@ -0,0 +1,124 @@
...
+# Here, we use [Harbor](https://www.harborframework.com/docs), a framework for evaluating and optimizing agents and models in container environments,
+# to demonstrate how to run massively parallel evals.
+
+import os
+import subprocess
+from pathlib import Path
+
+import modal
+
+app = modal.App("example-harbor")
```
charlesfrye, line 18: Style: name should be `example-harbor-evals`.

13_sandboxes/harbor_evals.py
```diff
@@ -0,0 +1,124 @@
...
+# to store generated artifacts
+
+
+volume = modal.Volume.from_name("example-harbor", create_if_missing=True)
+RUNS_PATH = Path("/runs")
+
+# harbor setup
+
+
+TASK_FILES = {
```
charlesfrye, line 40: Make this `DEFAULT_TASK_FILES` and add an option to pass task files through CLI/`local_entrypoint`.

charlesfrye (changes requested):

charlesfrye:
Discussed IRL -- better to demonstrate how to run `harbor` outside of Modal, with Modal as the backend, but that will require some ops work.

PR #1597, opened 2026-06-25
add computer use example

andrewhinh:
Add an example to demonstrate computer use over VNC, controlled by an LLM hosted on Modal AutoEndpoints.

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1597" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


13_sandboxes/computer_use_vnc.py
```diff
@@ -0,0 +1,270 @@
...
+    workspace.hydrate()
+    environment = os.environ.get("MODAL_ENVIRONMENT", "examples")
+    workspace_prefix = (
+        workspace.name
+        if environment in ("", "main")
+        else f"{workspace.name}-{environment}"
+    )
+    base_url = (
+        f"https://{workspace_prefix}--ep-{ENDPOINT_NAME}-server.us-west.modal.direct/v1"
+    )
```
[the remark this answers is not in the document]
andrewhinh: No current fix, discussed here: https://modal-com.slack.com/archives/C0AUXNZ8QS2/p1782346635401589

13_sandboxes/computer_use_vnc.py
```diff
@@ -0,0 +1,248 @@
...
+
+# ## Pinging the Endpoint
+#
+# When we create an Endpoint, it's considered "live" once it's finished provisioning.
+# However, it isn't necessarily ready to serve requests. The containers may have
+# scaled down to zero and need to be warmed. We also poll `/health` until it returns
+# HTTP 200, the same pattern we use for noVNC above.
+
+
+def is_endpoint_live() -> bool:
```
andrewhinh: Would rather do via SDK, discussed here: https://modal-com.slack.com/archives/C0AUXNZ8QS2/p1782346635401589

aaazzam (dismissed):
This example bundles two lessons into one file: serving your own model through an Endpoint, and driving a browser with an agent. The Endpoint setup takes up most of the page while the agent loop, which is really the point of the whole thing, get's less love.

I think zooming out there's a missing pedagogical bent. This feels more like a reference implementation than a walkthough, and I think the latter is more important for the docs. Loose opinions on some missing stuff:

- Opening (lines 6–12): it jumps straight into Endpoint and Sandbox mechanics before explaining what a "computer use" agent even is — that it looks at a screen and decides what to click, rather than calling a fixed set of APIs. That's the concept everything downstream depends on, and right now the reader is expected to already know it.
- Before the Endpoint setup (~line 50): there's no explanation for why you'd serve your own model at all instead of just pointing at a hosted API with a key. I don't think we need to go deep on this one, but like "you can use a hosted provider like OpenAI or Anthropic but for our purposes...".
- Before vnc_boot_command (~line 119): missing the reason there's a virtual display and VNC bridge in the first place ( without it, Browser Use would just run headless with nothing for you to watch).
- Before agent_command (~line 131), the section that currently has zero prose despite being the actual lesson: it's missing an explanation of the agent loop itself (look at the page, pick one action, repeat until it decides it's done), and it's missing why on_step_end exists at all — without that callback, nothing prints until the whole thing finishes.
- Before sandbox.create(...) (~line 198): missing the reasoning behind the two separate wait loops, one for the VNC feed and one for the Endpoint's health check, instead of just diving straight into running the agent. Starting before the model has warmed up would just produce a wall of connection errors.
- Missing entirely: a "Clean up" section. This one's more than a prose gap — the Endpoint this script creates has no teardown path at all. sandbox.terminate() only kills the Sandbox; the Endpoint stays deployed until someone happens to run modal endpoint stop themselves. I'd actually want this fixed in the script, not just called out in the docs.

Two smaller things for the same comment: agent_command gets built as a doubled-brace f-string, which is fragile to edit, and swapping in environment variables instead of string interpolation would let that whole block read like an ordinary file.

13_sandboxes/computer_use_vnc.py
```diff
@@ -0,0 +1,339 @@
+# ---
+# cmd: ["modal", "serve", "13_sandboxes/computer_use_vnc.py"]
```
charlesfrye: Testing should be done with `modal run`, exercise the example end-to-end by spinning up the agent and then hitting the `web_app`'s API for starting sessions, then checking for success.

charlesfrye (changes requested):

andrewhinh:
@prbot approve @charlesfrye

charlesfrye (approved):

PR #1598, opened 2026-06-25
Add: browser agent in a Modal Sandbox (Anthropic native bash tool + browse CLI)

shrey150:
## Summary

Adds **`13_sandboxes/browserbase_cli.py`**: a **browser agent** built on a Modal [Sandbox](https://modal.com/docs/guide/sandbox). The agent is an Anthropic tool-loop whose only tool is Claude's native **bash tool** (`{"type": "bash_20250124", "name": "bash"}`); each bash command runs inside the Sandbox via `sandbox.exec`, and the one useful program in the Sandbox is the Browserbase [`browse`](https://github.com/browserbase/stagehand/tree/main/packages/cli) CLI, which connects out over **CDP** to a remote browser running on [Browserbase](https://www.browserbase.com). The model drives the web by running `browse` commands; the heavy browser lives on Browserbase, so **no Chrome/Chromium is installed in the Sandbox**.

This example keeps the browser itself off the Sandbox entirely: Browserbase runs and maintains it, with a real fingerprint recognized as legitimate by bot protection partners and, on [Verified](https://www.browserbase.com/verified) sessions, automatic CAPTCHA solving. Wiring a browser into an existing agent harness becomes a CLI call away instead of a separate piece of infrastructure to build and operate.

### Architecture

```
┌─────────────────────┐   bash tool    ┌──────────────────┐  CDP over wss  ┌────────────────────────┐
│  Modal Function     │ ─────────────▶ │  Modal Sandbox   │ ─────────────▶ │  Browserbase browser   │
│  Claude agent loop  │  sandbox.exec  │  `browse` CLI    │ ◀───────────── │  (the real Chrome)     │
└─────────────────────┘ ◀───────────── └──────────────────┘   page data    └────────────────────────┘
```

The agent loop runs in a Modal Function. It creates a `modal.Sandbox` from an image built on `node:20-slim` with the `browse` CLI installed via `npm`, pinned to an explicit version, then runs each command the model emits via `sandbox.exec("bash", "-c", cmd)` and returns stdout/stderr as the `tool_result`. Design choices that keep it idiomatic and minimal:

- **Native bash tool, not a hand-rolled tool.** The model uses Claude's built-in bash tool — a schema-less Anthropic tool that's GA on Claude 4.x+ (no beta header needed). We don't define a custom `browse` tool or inject flags into the model's commands. The model is set to **`claude-sonnet-5`**.
- **We build the Sandbox image ourselves.** `modal.Image.from_registry("node:20-slim", add_python="3.12").run_commands(f"npm install -g browse@{BROWSE_VERSION} && npm cache clean --force")`, pinned to an explicit `BROWSE_VERSION` — matches Modal's stated preference for controlling image builds directly rather than depending on a third-party prebuilt image.
- **The Sandbox environment does the steering.** `BROWSERBASE_API_KEY` makes `browse` default to a remote Browserbase browser, and a `BROWSE_SESSION` generated fresh per invocation (`f"agent-{uuid.uuid4().hex[:8]}"`, not a shared literal) scopes every call to one shared session unique to that Sandbox, so the agent keeps the same browser tab across commands without any risk of colliding with another run.
- **Explicit long-lived Sandbox entrypoint.** A Sandbox lives only as long as its entrypoint process, so the example passes `"sleep", "infinity"` to `Sandbox.create` and drives the Sandbox purely with `sandbox.exec`, rather than relying on whatever the base image's own default command happens to do.
- **No separate `local_entrypoint`.** `run_agent` is the only Function in the file, so `modal run 13_sandboxes/browserbase_cli.py` invokes it directly and maps `--task` straight to the parameter — one less layer of indirection.
- **Prompt hardened for sandbox reality.** The system prompt is minimal — it enumerates no `browse` subcommands, just points the model at `browse skills show` (the CLI's bundled usage guide, shipped in `browse` ≥ 0.9.5) with `browse --help` as a supplementary reference. It also states the environment fact (Sandbox, no local browser, sessions remote and env-configured — never pass `--local`), injects **today's date at call time** (`build_system_prompt()`, `datetime.now(timezone.utc).date()`) so the model never guesses a date, and instructs the agent to stop navigating and give its best final answer when nearing the step limit (`MAX_STEPS = 40`).
- **Cleanup can't be skipped by a failed stop.** `browse stop` runs inside a `try`/`except` in the `finally` block so `sandbox.terminate()` always executes, even if the stop call itself fails (e.g. the Sandbox already timed out).
- **Timeouts sized off real ceilings, not round numbers.** The Function has an explicit `timeout=20 * MINUTES` (was relying on Modal's 5-minute default, tight for `MAX_STEPS=40`). Each `sandbox.exec` gets `1.5 * MINUTES` (90s) — Stagehand's own per-action default is 30s (`packages/core/lib/v3/understudy/page.ts`), so 90s covers a couple of chained `browse` calls with margin without masking a truly stuck command for minutes. Browserbase's own session timeout (project-default, up to 6 hours) is decoupled from all of this — every real run finishes in 1.5–3 minutes, so it's never the limiting factor here.

The default task is a **product-research** example that genuinely needs a browser: using Amazon, search for the current top mechanical keyboards and, for the top 5 results, compare each product's title, price, star rating, and number of ratings, returning a comparison table with each product's URL. Amazon's search renders its product grid client-side and returns nothing useful to a plain `curl`, so the agent must drive a real browser — exactly what `browse` provides. Override the goal with `--task "..."`.

The example uses a plain remote browser, which works on **any** Browserbase plan. (Verified browsers — residential IP + automatic CAPTCHA solving — are a Scale-plan upgrade; see https://www.browserbase.com/pricing.)

## Credentials & CI

Both the agent loop and the Sandbox need an Anthropic key and a Browserbase key. Following the sibling Anthropic examples (`sandbox_agent.py`, `anthropic_computer_use.py`), the example references a **named Modal Secret** rather than reading from the local env. Create it once before your first run:

```bash
modal secret create browser-agent ANTHROPIC_API_KEY=... BROWSERBASE_API_KEY=...
```

The Function and the Sandbox both take `modal.Secret.from_name("browser-agent", required_keys=["ANTHROPIC_API_KEY", "BROWSERBASE_API_KEY"])`.

Because the synthetic-monitoring system has no Browserbase secret, **the live monitoring run is disabled** via the frontmatter — the established pattern for examples that require a secret CI doesn't have:

```text
# lambda-test: false  # missing-secret
# pytest: false
```

The example still runs end-to-end with `modal run` once the `browser-agent` secret exists.

## Conformance to modal-examples conventions

- **Frontmatter:** `# lambda-test: false  # missing-secret`; `# pytest: false`.
- **Secrets:** `modal.Secret.from_name("browser-agent", required_keys=[...])`.
- **Images:** Sandbox image built from `node:20-slim` (`add_python="3.12"`) with `browse` installed via `npm`, pinned to an explicit version; Function image `debian_slim(python_version="3.12")` with pinned `anthropic==0.115.0`.
- **Timeouts:** `MINUTES = 60` module constant; Sandbox timeout `15 * MINUTES`, Function timeout `20 * MINUTES`, per-command `sandbox.exec` timeout `1.5 * MINUTES`.
- **Naming:** App id `example-browserbase-cli`; no separate `local_entrypoint` since there's a single Function.
- `ruff check` + `ruff format` clean.

## E2E Test Matrix

Run against **released `browse` 0.9.5** on real Modal cloud, executing the example exactly as committed on this branch.

| Command / flow | Observed output | Confidence / sufficiency |
| --- | --- | --- |
| `modal run 13_sandboxes/browserbase_cli.py` (full agent run, default Amazon task, `claude-sonnet-5`, native bash tool) | Image builds fresh from `node:20-slim` + `npm install -g browse@0.9.5` in ~14s. **First tool call was `browse skills show`.** Agent completed with `===== FINAL ANSWER =====` and a correct 5-row comparison table (title, price, star rating, # of ratings, canonical `amazon.com/dp/<ASIN>` URL per row), sponsored placements excluded, clean `browse stop` + `sandbox.terminate()`, app reached `completed` state. | Proves the example runs end-to-end as written on the current architecture: fresh npm-built image, unique per-run session, auto-detected entrypoint, hardened cleanup. Run multiple times across the changes in this PR with consistent results. |
| Same run, repeated after removing `local_entrypoint()` | `modal run` auto-detected `run_agent` as the sole Function and mapped `--task` to it directly, with no `local_entrypoint` defined — confirmed against [Modal's docs](https://modal.com/docs/guide/apps): "If you have no entrypoint specified, and just one decorated Modal Function, that will be used as a remote entrypoint instead." | Confirms removing the wrapper function is behavior-preserving, not just theoretically per the docs but by an actual passing run. |
| Same run, repeated after the file rename (`browsecli_in_modal.py` → `browserbase_cli.py`) and after switching `BROWSE_SESSION` to a per-invocation generated value | Both completed cleanly with the same output shape as the baseline run. | Confirms the rename and the session-uniqueness fix didn't change runtime behavior. |
| Regression check after wrapping `browse stop` in `try`/`except` in the `finally` block | Happy-path run still completes normally with the same output shape. | Confirms the cleanup hardening doesn't affect the non-failure path (the failure path itself — Sandbox already gone when `stop` is attempted — is a narrow race that's impractical to force organically; the fix is a standard try/except so `sandbox.terminate()` is unconditionally reached). |
| Historical regression that motivated the entrypoint fix: run **without** `"sleep", "infinity"` | Every `sandbox.exec` failed with `NotFoundError: Task has already finished with status success` — a Sandbox with no long-lived entrypoint process exits almost immediately, taking the Sandbox with it. | Demonstrates the failure mode the explicit no-op entrypoint fixes; the fixed runs above are the proof it works. |
| Cleanup: Modal app state + Browserbase sessions API `status=RUNNING` after each run above | App transitions to `completed`/`stopped`, `[]` for running Browserbase sessions. | Confirms no leaked Sandboxes or remote browser sessions across the runs above. |
| Regression check after widening timeouts (Function `20 * MINUTES`, per-command `1.5 * MINUTES`) | Happy-path run still completes normally with the same output shape, well under either ceiling. | Confirms the wider timeouts don't change observed behavior on a normal run; sizing rationale (Stagehand's own 30s per-action default) is in the Architecture section above. |
| `ruff check` + `ruff format --check` on the committed file | All checks passed; formatting clean. | Supporting: file conforms to repo lint/format conventions. |

## Type of Change

  - [ ] New example for the documentation site

## Monitoring Checklist

- [x] Example is configured with `cmd: ["modal", "run", "13_sandboxes/browserbase_cli.py"]`
  - [x] Tested by executing with `modal run` on real Modal cloud (with the `browser-agent` secret) against released `browse` 0.9.5
  - [x] Live synthetic-monitoring run is intentionally disabled via `# lambda-test: false  # missing-secret` (CI has no Browserbase secret)
  - [x] Runs with no arguments (optional `--task` overrides the research goal)
  - [x] Function image deps pinned (`anthropic==0.115.0`); the Sandbox image installs `browse` pinned to an explicit version at build time.

## Notes for maintainers

- **Requires a named Modal Secret** `browser-agent` with `ANTHROPIC_API_KEY` and `BROWSERBASE_API_KEY` (matching `sandbox_agent.py` / `anthropic_computer_use.py`, which use `anthropic-secret`).
- Single self-contained file; docs are embedded as header comments in the `.py` (matching the gallery's literate-example style), following the `13_sandboxes/` Sandbox examples (`safe_code_execution.py`, `sandbox_agent.py`).
- The Sandbox image is built from `node:20-slim` with `browse` installed via `npm install -g browse@<pinned version>` — we control the image build directly rather than depending on a third-party prebuilt image.


shrey150:
Updated: the agent's system prompt now points it at `browse skills show` first — the CLI's bundled usage guide — with `browse --help` kept as a supplementary reference. `browse skills show` ships in the next `browse` release; the Sandbox image already tracks `ghcr.io/browserbase/browse`, so it'll pick it up automatically once released.

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
+# ---
+# cmd: ["modal", "run", "13_sandboxes/browsecli_in_modal.py"]
+# lambda-test: false  # missing-secret
+# pytest: false
+# ---
+
+# # Run a browser agent in a Modal Sandbox
+
+# A Modal [Sandbox](https://modal.com/docs/guide/sandbox) is a great place to run an
```
charlesfrye, line 9: I would rather not frame the intro to this around weakness of Modal Sandboxes, in favor of framing in terms of strengths of Browserbase.

In particular, all of the features you describe could be built on Modal Sandboxes (e.g. using proxies to solve IP issues), and we don't want to imply that it's not possible.

Also, our Sandboxes are not Firefracker microVMs.

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+# The agent's task (overridable): research the current top mechanical keyboards on
+# Amazon and return a comparison of each product's title, price, star rating, and
+# number of ratings. Amazon search renders its product grid client-side and returns
+# nothing useful to a plain `curl`, so the agent has to drive a real browser — which
+# is exactly what `browse` gives it.
+
+# To run it:
+
+# ```bash
+# export ANTHROPIC_API_KEY=sk-ant-...
```
charlesfrye: This doesn't seem right. The secret values are only needed on Modal, correct?

13_sandboxes/browserbase_cli.py
charlesfrye, line 1: I would prefer a different name here. `browsecli` maybe? Or `browserbase_cli`. Drop the `modal`, go for something a bit more generic.

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+import modal
+
+# ## Build the Sandbox image
+#
+# We pull the official prebuilt `ghcr.io/browserbase/browse` image, which is
+# `node:20-slim` with the `browse` CLI already installed — so there's no inline
+# `npm install` step. You can pin a version with a tag (e.g.
+# `ghcr.io/browserbase/browse:0.9.4`). **No Chrome/Chromium is bundled** — the browser
```
charlesfrye: I don't think we need to repeat ourselves.

charlesfrye: Is there a strong reason not to just run `npm install` via `run_commands`? We generally prefer to control Image builds as much as possible, and it's what we recommend to our customers.

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+# ```
+
+
+import modal
+
+# ## Build the Sandbox image
+#
+# We pull the official prebuilt `ghcr.io/browserbase/browse` image, which is
+# `node:20-slim` with the `browse` CLI already installed — so there's no inline
+# `npm install` step. You can pin a version with a tag (e.g.
```
charlesfrye: Please pin the version so that we can be confident the example will continue to work for new users as versions change.

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# lives on Browserbase and is reached over CDP at run time. The agent loop itself runs
+# in a Modal Function and only needs `anthropic`, so we add that to the Function's image
```
charlesfrye: Don't bring this up yet.

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+sandbox_image = modal.Image.from_registry(
+    "ghcr.io/browserbase/browse", add_python="3.12"
+)
+
+app = modal.App("example-browsecli-in-modal")
+
+MINUTES = 60  # seconds, for readable timeouts
+
+# ## Credentials
+#
```
charlesfrye:
```suggestion

```

Part of the style in this repo.

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# Both the agent loop and the Sandbox need an Anthropic key and a Browserbase key.
+# We store them once as a named Modal Secret and reference it by name. Create it
+# before your first run with:
+#
```
charlesfrye:
```suggestion

```

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+
+app = modal.App("example-browsecli-in-modal")
+
+MINUTES = 60  # seconds, for readable timeouts
+
+# ## Credentials
+#
+# Both the agent loop and the Sandbox need an Anthropic key and a Browserbase key.
+# We store them once as a named Modal Secret and reference it by name. Create it
+# before your first run with:
```
charlesfrye, line 71: You can also direct to the dashboard, https://modal.com/secrets

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# ```bash
+# modal secret create browser-agent ANTHROPIC_API_KEY=... BROWSERBASE_API_KEY=...
+# ```
+
+agent_secret = modal.Secret.from_name(
+    "browser-agent", required_keys=["ANTHROPIC_API_KEY", "BROWSERBASE_API_KEY"]
+)
+
+# ## The agent
+#
```
charlesfrye:
```suggestion

```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# The model's only tool is Claude's built-in **bash tool**
+# (`{"type": "bash_20250124", "name": "bash"}`) — a schema-less tool whose commands we
+# execute ourselves. We run each command inside the Sandbox with `sandbox.exec`, so the
+# model gets a real shell whose one useful program is `browse`.
+#
```
charlesfrye:
```suggestion

```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# We don't hand-roll a custom `browse` tool or inject flags into the model's commands.
+# Instead the Sandbox's environment does the steering: `BROWSERBASE_API_KEY` makes
+# `browse` default to a **remote** Browserbase browser, and `BROWSE_SESSION=agent`
+# scopes every call to one shared session, so the agent keeps the same browser tab
+# across commands.
+#
```
charlesfrye:
```suggestion

```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+
+sandbox_image = modal.Image.from_registry(
+    "ghcr.io/browserbase/browse", add_python="3.12"
+)
+
+app = modal.App("example-browsecli-in-modal")
+
+MINUTES = 60  # seconds, for readable timeouts
+
+# ## Credentials
```
charlesfrye: Strong preference for verb phrases -- something like "Manage credentials" or "Add credentials to a Modal Secret"

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+DEFAULT_TASK = (
+    "Using Amazon (https://www.amazon.com), research the current top mechanical keyboards: "
+    "search the site, then for the top 5 results compare each product's title, price, star "
+    "rating, and number of ratings. Return a comparison table including each product's URL."
+)
+MODEL = "claude-sonnet-5"
+MAX_STEPS = 40
+MAX_OUTPUT_CHARS = 40_000  # cap each tool result so it fits the context budget
+
+SYSTEM_PROMPT = "You are an autonomous deep-research agent. You have a `browse` CLI (Browserbase browser automation) in your bash tool — it is installed, and its auth and a shared browser session are already configured via environment variables. Learn how to use it by running `browse skills show` (it prints the CLI's bundled usage guide); use `browse --help` (and `browse <command> --help`) for extra detail on any command. Then complete the task. When you cite a document, link the direct document itself, not a viewer, preview, or index page that wraps it. Return a clear, well-sourced answer."
```
charlesfrye: Could we break this up onto multiple lines, as in the `DEFAULT_TASK`?

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+    "rating, and number of ratings. Return a comparison table including each product's URL."
+)
+MODEL = "claude-sonnet-5"
+MAX_STEPS = 40
+MAX_OUTPUT_CHARS = 40_000  # cap each tool result so it fits the context budget
+
+SYSTEM_PROMPT = "You are an autonomous deep-research agent. You have a `browse` CLI (Browserbase browser automation) in your bash tool — it is installed, and its auth and a shared browser session are already configured via environment variables. Learn how to use it by running `browse skills show` (it prints the CLI's bundled usage guide); use `browse --help` (and `browse <command> --help`) for extra detail on any command. Then complete the task. When you cite a document, link the direct document itself, not a viewer, preview, or index page that wraps it. Return a clear, well-sourced answer."
+
+# Claude's native bash tool. The schema is built into the model — we declare it by
+# type and name only, and our handler executes the `command` Claude sends.
```
charlesfrye, line 137:
```suggestion
# type and name only, and our handler executes the `command` Claude sends.

```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+
+# Claude's native bash tool. The schema is built into the model — we declare it by
+# type and name only, and our handler executes the `command` Claude sends.
+BASH_TOOL = {"type": "bash_20250124", "name": "bash"}
+
+
+@app.function(
+    image=modal.Image.debian_slim(python_version="3.12").pip_install(
```
charlesfrye: I think it's better to create this as a separate step.

The two Images could both be created together in a section on container Images or they could be created separately just before use.

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+        "anthropic==0.115.0"
+    ),
+    secrets=[agent_secret],
+)
+def run_agent(task: str = DEFAULT_TASK) -> str:
+    import anthropic
+
+    # Boot the Sandbox that holds the `browse` CLI. The environment steers every
+    # command to a shared *remote* browser, so the model never has to pass flags.
```
charlesfrye: This comment feels excessive.

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+# The model's only tool is Claude's built-in **bash tool**
+# (`{"type": "bash_20250124", "name": "bash"}`) — a schema-less tool whose commands we
+# execute ourselves. We run each command inside the Sandbox with `sandbox.exec`, so the
+# model gets a real shell whose one useful program is `browse`.
+#
+# We don't hand-roll a custom `browse` tool or inject flags into the model's commands.
+# Instead the Sandbox's environment does the steering: `BROWSERBASE_API_KEY` makes
+# `browse` default to a **remote** Browserbase browser, and `BROWSE_SESSION=agent`
+# scopes every call to one shared session, so the agent keeps the same browser tab
+# across commands.
```
charlesfrye: Isn't this also going to mean the browser is shared across executions in the same environment? Maybe it should instead be an ID based on the sandbox's `object_id`?

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,211 @@
...
+#
+# The system prompt is deliberately minimal: it doesn't enumerate any `browse`
+# subcommands. It just tells the model the CLI is installed and pre-configured and
+# points it at `browse skills show` (the CLI's bundled usage guide), with
+# `browse --help` as a supplementary reference, to learn the interface itself, then
+# lets it plan its own steps for whatever task it's given.
+
+SESSION = "agent"
```
charlesfrye:
```suggestion
BROWSE_SESSION = "agent"
```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+        run_bash("browse stop")
+        sandbox.terminate()
+
+    answer = final_text or "(no answer)"
+    print("\n===== FINAL ANSWER =====\n" + answer)
+    return answer
+
+
+# ## Local entrypoint
+#
```
charlesfrye:
```suggestion

```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+
+@app.local_entrypoint()
+def main(task: str = DEFAULT_TASK):
+    run_agent.remote(task)
+
+
+# > **Note on protected sites.** This example uses a plain remote browser, which works
+# > on **any** Browserbase plan. To reach sites behind aggressive bot-detection,
+# > Browserbase also offers Verified browsers (residential IP + automatic CAPTCHA
+# > solving), which require a Scale plan — see https://www.browserbase.com/pricing.
```
charlesfrye:
```suggestion
# > solving), which require [a Scale plan](https://www.browserbase.com/pricing).
```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+# Function in the cloud. Pass a different goal with `--task "..."`.
+
+
+@app.local_entrypoint()
+def main(task: str = DEFAULT_TASK):
+    run_agent.remote(task)
+
+
+# > **Note on protected sites.** This example uses a plain remote browser, which works
+# > on **any** Browserbase plan. To reach sites behind aggressive bot-detection,
```
charlesfrye:
```suggestion
# > on any Browserbase plan. To reach sites behind aggressive bot-detection,
```

13_sandboxes/browsecli_in_modal.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+# ## Local entrypoint
+#
+# `modal run 13_sandboxes/browsecli_in_modal.py` triggers this, which runs the agent
+# Function in the cloud. Pass a different goal with `--task "..."`.
+
+
+@app.local_entrypoint()
+def main(task: str = DEFAULT_TASK):
```
charlesfrye: You actually don't need this -- you can invoke `run_agent` directly, including passing CLI arguments.

charlesfrye (changes requested):
HI @shrey150!

Thanks for sharing this sample code. We are interested in merging, subject to some revisions -- and I will follow up with some adjustments to get the code to work in our QA system.

FYI, Modal and Browserbase have a shared Slack channel. Feel free to join that and then ping me there or in DM with questions.

In addition to review comments, note that the mypy tests are failing.

shrey150:
browse@0.9.5 is now released (npm + `ghcr.io/browserbase/browse:latest`), including the `browse skills show` command this example's prompt relies on — so the example now runs end-to-end as written.

Verified on real Modal cloud against the released image: `browse --version` → 0.9.5 inside a Sandbox built from the committed image definition, then a full `modal run` of the default Amazon task — the agent's first tool call was `browse skills show`, it finished in 8 tool calls (~90s) with zero errors and returned the full 5-product comparison table.

One real fix surfaced by this run: the official image's default CMD is `browse --help`, which exits immediately, and a Modal Sandbox lives only as long as its entrypoint — so `Sandbox.create` now passes an explicit `"sleep", "infinity"` no-op entrypoint (otherwise every `sandbox.exec` fails with 'Task has already finished with status success'). E2E matrix in the PR body updated with these runs; marking ready for review.

charlesfrye:
@shrey150 I am assuming that last comment was by your agent. Please show up yourself in this thread (or in Slack) if you want this to be merged.

13_sandboxes/browsecli_in_modal.py
shrey150: no strong reason - can swap to this approach!

13_sandboxes/browserbase_cli.py
```diff
@@ -0,0 +1,248 @@
...
+    # long-lived no-op entrypoint and drive it entirely with `sandbox.exec`.
+    sandbox = modal.Sandbox.create(
+        "sleep",
+        "infinity",
+        app=app,
+        image=sandbox_image,
+        secrets=[agent_secret],
+        env={"BROWSE_SESSION": browse_session},
+        timeout=15 * MINUTES,
+    )
```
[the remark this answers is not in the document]
shrey150: project ID is not needed

charlesfrye:
Hello!

PR #1604, opened 2026-07-08
Add Anthropic Claude batch inference example with structured output

steps-re:
Adds an `llm-serving` example for running Claude batch inference on Modal.

What it shows:
- Concurrent batch inference by fanning prompts out across workers with `extract_entities.map()`
- Typed structured output via Instructor + Pydantic — Claude returns a validated `EntityExtractionResult`, not raw text
- A FastAPI gateway exposed with `@modal.asgi_app()`, with a `POST /extract` route protected by an `X-API-Key` header checked against a Modal secret

`modal run` fans the sample passages out in parallel and prints the results; `modal deploy` stands up the API. Setup needs two secrets — `anthropic-secret` and `batch-api-key` — documented in the header.

Follows the `06_gpu_and_ml/llm-serving` conventions (`pytest: false` frontmatter, literate comments, `with image.imports()` for runtime deps, secrets pinned with `required_keys`). Passes `ruff`.

Verified by running `modal run` end to end: it extracts entities from the sample passages in parallel through Claude and returns structured output.

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1604" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/06_gpu_and_ml/llm-serving/anthropic_batch_inference.py
+++ b/06_gpu_and_ml/llm-serving/anthropic_batch_inference.py
@@ -64,10 +64,10 @@ from pydantic import BaseModel, Field
 app = modal.App("example-anthropic-batch-inference")
 
 image = modal.Image.debian_slim(python_version="3.11").pip_install(
-    "anthropic>=0.50.0",
-    "instructor>=1.7.0",
-    "fastapi[standard]",
-    "pydantic>=2.0",
+    "anthropic==0.116.0",
+    "instructor~=1.15.0",
+    "fastapi[standard]==0.139.0",
+    "pydantic~=2.13.0",
 )
 
 # ## Pydantic models for structured output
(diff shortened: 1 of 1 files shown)


steps-re:
Thanks for the review. Addressed both points:

- Pinned all container dependencies per the repo convention: `anthropic==0.116.0`, `fastapi[standard]==0.139.0` (exact patch for < 1.0), and `instructor~=1.15.0`, `pydantic~=2.13.0`.
- Pinned the model to the exact snapshot `claude-haiku-4-5-20251001` for reproducibility. I'd already verified the example end to end with `modal run` (it resolves and returns structured extractions); the dated snapshot just makes that deterministic.

charlesfrye:
Hey there! Thanks for sharing your code. I don't think we're interested in this particular use case.

PR #1605, opened 2026-07-08
Clairec03/modal server example

clairec03:
## Type of Change

## Monitoring Checklist

### Content

### Build Stability

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1605" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,259 @@
...
+# It polls `/health` until a container is ready,
+# retrying the 503s that cold starts produce,
+# then requests one document embedding and verifies its shape.
+
+
+@app.local_entrypoint()
+def main(timeout_s: float = 600):
+    import time
+
+    import requests
```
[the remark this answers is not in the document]
clairec03: Already removed 3rd-party library in a recent commit

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,273 @@
...
+# We persist the llama.cpp download cache in a Modal
+# [Volume](https://modal.com/docs/guide/volumes)
+# so the GGUF file is downloaded from the Hub exactly once
+# and loaded from the Volume on later cold starts.
+
+CACHE_PATH = "/cache"
+
+volume = modal.Volume.from_name(
+    "liquidai-embeddings-cache", create_if_missing=True, version=2
+)
```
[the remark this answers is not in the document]
clairec03: Addressed

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,273 @@
...
+
+# ## Choose the model file and engine parameters
+
+# Liquid AI publishes official GGUF conversions of the model in
+# [LiquidAI/LFM2.5-Embedding-350M-GGUF](https://huggingface.co/LiquidAI/LFM2.5-Embedding-350M-GGUF).
+# We serve the F16 file, a near-lossless conversion of the BF16 training precision.
+# At roughly 700 MB, the file is already small enough to not require quantization.
+# `llama-server` downloads the file from the Hugging Face Hub on first start.
+
+MODEL_GGUF = "LiquidAI/LFM2.5-Embedding-350M-GGUF:F16"
```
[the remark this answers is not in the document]
clairec03: Addressed

06_gpu_and_ml/embeddings/liquidai_embeddings_client.py
```diff
@@ -0,0 +1,116 @@
...
+
+# We look up the deployed Server's URL by name,
+# wait for a container to be ready, and embed one document.
+
+if __name__ == "__main__":
+    text = "The quick brown fox jumps over the lazy dog."
+
+    try:
+        server = modal.Server.from_name(APP_NAME, SERVER_NAME)
+        url = server.get_url()
```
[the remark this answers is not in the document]
clairec03: Addressed

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,284 @@
...
+# ## Define the container image
+
+# We build on the official llama.cpp server image.
+# It contains the compiled binary and doesn't include Python,
+# so `add_python` bundles an interpreter for Modal's own runtime.
+# We also clear the image's entrypoint, which is the server binary itself,
+# because we launch that binary ourselves in the server's startup hook.
+
+image = (
+    modal.Image.from_registry("ghcr.io/ggml-org/llama.cpp:server", add_python="3.12")
```
akelch11: we want to pin image versions to keep things stable. In this case I think you can add -b9932 which is the [latest release tag](https://github.com/ggml-org/llama.cpp/releases/tag/b9932)

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,284 @@
...
+                str(N_SLOTS),
+                "--ctx-size",
+                str(N_CTX),
+                "--batch-size",
+                str(N_CTX),
+                "--ubatch-size",
+                str(N_CTX),
+            ],
+            start_new_session=True,
+        )
```
akelch11: Can we add a health check loop in the modal.enter()?

That way by the time the container is only considered "live" once the server is ready.

This pattern is in our other inference server examples, see 

- https://modal.com/docs/examples/lfm_snapshot#low-latency-serverless-lfm2-with-vllm-and-modal
- https://modal.com/docs/examples/amazon_embeddings#embed-30-million-amazon-reviews-at-575k-tokens-per-second-with-qwen2-7b

06_gpu_and_ml/embeddings/liquidai_embeddings_client.py
```diff
@@ -0,0 +1,115 @@
+# ---
```
akelch11: our philosophy on examples is that they are "radically single file". They should fit in a single python file and web page for dev simplicity. This client is mostly just a wrapper around sending OpenAI compatible requests and we can just leave the at the end of the server file / doc as you did. Let's remove this

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,284 @@
...
+
+# `llama-server` processes requests in `N_SLOTS` parallel slots
+# and splits the total token context evenly across them.
+# We give each slot exactly the model's trained sequence length of 512 tokens.
+# In embedding mode, llama.cpp requires each input to fit in a single physical batch,
+# and caps the logical batch size at the physical batch size.
+# Sizing both batches to the full context lets all `N_SLOTS` slots
+# process maximal inputs in one forward pass.
+
+MAX_INPUT_TOKENS = 512  # the model's trained sequence length
```
akelch11, line 65: our philosophy on constants is to keep them minimal, and somewhat counter intuitively to normal programming, introduce them as needed to make reading easier

from[ style manifesto](https://app.notion.com/p/modal-com/Examples-Style-Manifesto-2261e7f1694980ae86c8fbfd9a10be61?source=copy_link#2271e7f1694980209367f71e9706a265):
```
So when writing, you want to **minimize the gap between time-of-creation and time-of-use** for objects and **minimize the gap between first reference to a concept and first use**.

This is often quite different to the way code is written normally. For instance, in the examples we often defer “magic constants” to the site of usage, rather than keeping them all at the top of the file, where they are likely to introduce a large number of concepts at once, long before they are used.
```

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,284 @@
...
+@app.server(
+    image=image,
+    volumes={CACHE_PATH: volume},
+    port=PORT,
+    cpu=CPU_COUNT,
+    memory=MEMORY_MB,
+    target_concurrency=TARGET_CONCURRENCY,
+    min_containers=MIN_CONTAINERS,
+    startup_timeout=10 * MINUTES,  # first-ever start downloads the GGUF
+    scaledown_window=5 * MINUTES,  # idle containers shut down after 5 minutes
```
akelch11: nit: some of these comments aren't needed, such as unauth, and scaledown window

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,284 @@
...
+# That makes Servers a natural fit for llama.cpp.
+# The `llama-server` binary uses HTTP natively and implements the
+# [OpenAI embeddings API](https://platform.openai.com/docs/api-reference/embeddings)
+# out of the box.
+# Support for the LFM2.5 embedding model landed in
+# [ggml-org/llama.cpp#24913](https://github.com/ggml-org/llama.cpp/pull/24913).
+# Since `llama-server` handles the requests itself, the Python code in this
+# example only starts and stops that process, and no web framework is needed.
+
+# Note that Modal Server does not queue requests.
```
akelch11: We prefer to[ keep detailed explanations of features/components in one place](https://app.notion.com/p/modal-com/Examples-Style-Manifesto-2261e7f1694980ae86c8fbfd9a10be61?source=copy_link#2261e7f1694980fe8018d7f3cb583a99), the formal reference guide. 

I'd compare your discussion of app.server to another inference example (e.g, [this](https://modal.com/docs/examples/lfm_snapshot#low-latency-serverless-lfm2-with-vllm-and-modal))and keep it consistent (all the app.server inference examples basically copy paste explanations, and we don't want this to be the outlier where details come from, let's keep that to the formal docs)

but this is a good thing to know/learn so it wasn 't in vain!

akelch11 (changes requested):
Will pick this back up tomorrow in more detail but overall looks really good and it's good to have a llama.cpp thing up! Wanted to get you some timely feedback so may revisit a 2nd round

Nice job keeping to stdlib + modal. Here's some extra resources on adding examples that may be useful to look at on another pass

[Examples FDE Lab](https://docs.google.com/presentation/d/1X7gRiA11vQTZGXuJG6ZA9Y8_gPdmC2OXQa3ZS55tmO4/edit?slide=id.g3d233bec54e_0_381#slide=id.g3d233bec54e_0_381)
[Examples style manifesto](url)

06_gpu_and_ml/embeddings/liquidai_embeddings_client.py
clairec03: Addressed

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Addressed

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,259 @@
+# ---
+# deploy: true
+# cmd: ["modal", "run", "06_gpu_and_ml/embeddings/liquidai_embeddings_server.py"]
+# ---
+
+# # Serve Liquid AI embeddings with Modal Servers
```
akelch11: let's mention llama.cpp in the title, how about "Serve Liquid AI embeddings with llama.cpp and Modal Servers

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@app.server(
+    image=image,
+    volumes={CACHE_PATH: volume},
+    port=PORT,
+    cpu=N_SLOTS,
+    memory=2048,  # MBs
+    target_concurrency=N_SLOTS,
+    min_containers=0,  # change to 1 or more for production
```
akelch11: let's say
"# set to 1 or more to ensure a replica is always ready"

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,259 @@
...
+# It is also small enough to serve economically on CPU, so this example doesn't require a GPU.
+
+# For client code, use the appropriate prompt prefixes for inputs.
+# Prepend `"query: "` when embedding search queries and `"document: "` when embedding passages.
+# Omitting the prefixes silently degrades retrieval quality, since the model was trained as
+# an asymmetric retriever.
+
+# ## Why use a Modal Server?
+
+# [Modal Servers](https://modal.com/docs/guide/servers) route requests directly
```
akelch11: Can we include this standard blurb about servers being good for low latency inference in here

"
To minimize routing overheads, we use @app.server, which uses a new, low-latency routing service on Modal designed for latency-sensitive inference workloads. 
"
from https://modal.com/docs/examples/sglang_low_latency#low-latency-qwen-36-with-sglang-and-modal and others

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,259 @@
...
+# In this example, we serve
+# [LiquidAI/LFM2.5-Embedding-350M](https://huggingface.co/LiquidAI/LFM2.5-Embedding-350M)
+# behind an OpenAI-compatible `POST /v1/embeddings` API
+# using [llama.cpp](https://github.com/ggml-org/llama.cpp)
+# and [Modal Servers](https://modal.com/docs/guide/servers).
+
+# LFM2.5-Embedding-350M is a 350M-parameter multilingual embedding model.
+# It is a dense bidirectional encoder that produces one 1024-dimensional vector per input
+# and supports retrieval across eleven languages.
+# It is also small enough to serve economically on CPU, so this example doesn't require a GPU.
```
akelch11, line 17: just curious, do you have an idea of the latency perf here on CPU? 350M feels like something that could still use t4

akelch11: i guess the benchmarks are on a Macbook LOL fair enough

akelch11 (approved):
Looks good to me now, thanks for revising! When the Colbert multivector version hits huggingface it would be cool to see that too

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
+# ---
+# deploy: true
```
charlesfrye, line 2: Why are we deploying this?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# cmd: ["modal", "run", "06_gpu_and_ml/embeddings/liquidai_embeddings_server.py"]
```
charlesfrye, line 3: This is not needed, it is the default value for `cmd`

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# ---
+
+# # Serve Liquid AI embeddings with llama.cpp and Modal Servers
```
charlesfrye, line 6: Why are we serving with llama.cpp?

charlesfrye: I would expect vLLM and SGLang to perform better. Do we have benchmarks to indicate otherwise?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# For client code, use the appropriate prompt prefixes for inputs.
+# Prepend `"query: "` when embedding search queries and `"document: "` when embedding passages.
+# Omitting the prefixes silently degrades retrieval quality, since the model was trained as
+# an asymmetric retriever.
+
+# ## Why use a Modal Server?
+
+# To minimize routing overheads, we use `@app.server`,
+# which uses a new, low-latency routing service on Modal designed for latency-sensitive inference workloads.
+# See [the reference documentation](https://modal.com/docs/reference/modal.App#server) for details.
```
charlesfrye, line 28: Prefer Guides over references when linking from examples.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+
+# The `llama-server` binary speaks HTTP natively and implements the
+# [OpenAI embeddings API](https://platform.openai.com/docs/api-reference/embeddings)
+# out of the box, so the Python code in this example only starts and stops
+# that process, with no web framework needed.
+# Support for the LFM2.5 embedding model landed in
```
charlesfrye, line 34: Deltte this line.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+import urllib.error
+import urllib.request
+
+import modal
+
+# ## Choose the model file and engine parameters
+
+# Liquid AI publishes official GGUF conversions of the model in
+# [LiquidAI/LFM2.5-Embedding-350M-GGUF](https://huggingface.co/LiquidAI/LFM2.5-Embedding-350M-GGUF).
+# We serve the F16 file, a near-lossless conversion of the BF16 training precision.
```
charlesfrye, line 48: Why?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# At roughly 700 MB, the file is already small enough to not require quantization.
```
charlesfrye, line 49: What is small here? How are requirements determined?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# `llama-server` downloads the file from the Hugging Face Hub on first start.
```
charlesfrye, line 50: Why not on subsequent starts?

You are getting ahead of yourself. The material on cacheing with Modal Volumes is later. Remove this sentence.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+)
+
+# ## Define the Server
+
+# We wrap the engine in a class registered with `@app.server()`,
+# which attaches the image, Volume, and resources
+# and fronts the containers with a proxy.
+# The `@modal.enter` and `@modal.exit` lifecycle hooks below
+# start and stop the `llama-server` process.
+# See the [reference documentation](https://modal.com/docs/reference/modal.App#server) for details.
```
charlesfrye, line 104: Prefer guides over references when linking from examples.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+            time.sleep(min(delay, remaining))
+            delay = min(delay * 2, 10.0)
+    raise TimeoutError(
+        f"Liquid AI embeddings server not ready within {timeout} seconds"
+    )
+
+
+app = modal.App("example-liquidai-embeddings")
+
+# The resource reservations follow from the engine parameters:
```
charlesfrye, line 136: On Modal, you don't need to reserve the entirety of the resources you consume at peak. Did you look at the consumed cores and memory to determine whether this is needed?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+# 1 CPU core per slot, and 2 GB of memory to hold
+# the roughly 700 MB of F16 weights plus KV cache and other overhead.
+# Setting `target_concurrency` to the slot count sends each container
+# only as many concurrent requests as it has slots,
+# and load beyond that scales up new containers instead
```
charlesfrye, line 141: Not true for Servers -- it is `target_concurrency`, not `max_concurrency`. This is an important detail to get right in a guide for Servers.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+    volumes={CACHE_PATH: volume},
+    port=PORT,
+    cpu=N_SLOTS,
+    memory=2048,  # MBs
+    target_concurrency=N_SLOTS,
+    min_containers=0,  # set to 1 or more to ensure a replica is always ready
+    startup_timeout=10 * MINUTES,  # allows time to download the GGUF on startup
+    scaledown_window=5 * MINUTES,
+    exit_grace_period=20,
+    unauthenticated=True,
```
charlesfrye, line 156: Comment on this.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+)
+class LlamaCppEmbeddingServer:
+    @modal.enter()
+    def start(self):
+        # The image sets LLAMA_ARG_HOST=0.0.0.0, which binds all interfaces,
```
charlesfrye, line 161: Do we need to say this here? Or alternatively, do we need to use an env var when we could be explicit?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
```diff
@@ -0,0 +1,260 @@
...
+        # so we do not pass --host.
+        # start_new_session blocks llama-server from receiving the container-wide
```
charlesfrye, line 163: move this commentary down to the place where it applies, ie on the Popen, not on the cmd

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
charlesfrye: `deploy: true` is for examples we want up as demo deployments. That does not include this one.

charlesfrye:
hey @clairec03 some followups in review

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: This is not supported by vLLM and SGLang yet.

charlesfrye: What is "this"? This model?

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: What's the suggestion here? Should I delete this comment or switch models??

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: LFM 2.5 retrievers were designed for portable deployment (small and can "run nearly anywhere (CPUs, laptops, and edge devices)"), as documented on Liquid AI's own [website](https://www.liquid.ai/blog/lfm2-5-retrievers).

charlesfrye: My point is that you need to quantify these claims. The writing here is closer to LiquidAI's marketing copy, not an explanation for an engineer.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
charlesfrye: If we're going to call this out, we should explain _why_. Why are we running the fp16 version instead of the bf16 version?

Most format changes are for faster inference, but I don't think that's what's happening here.

I would guess that the reason why is that we are running on a CPU here, rather than a GPU, and so there is not native support for bf16 arithmetic. That's a notable feature of this example, actually, because "typical" ML deployments these days are on GPUs.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Yes

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: I meant to say that the download only happens once for each new container during startup.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Removed resource (memory and CPU) reservation.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: SGLang doesn't support Liquid AI models (see [details](https://sgl-project-sglang-93.mintlify.app/models/supported-models#model-specific-documentation)), and vLLM hasn't added support for bidirectional embeddings in LFM2.5 - only dense, MoE, and VL models (see [details](https://docs.liquid.ai/deployment/gpu-inference/vllm)).

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Great suggestion! I looked at the consumed cores (3.36) and memory (106.48 MB) which are both small. So I think this is probably not needed.

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Revised the language to make it clearer and removed the claim about file size

06_gpu_and_ml/embeddings/liquidai_embeddings_server.py
clairec03: Switched to BF16

PR #1612, opened 2026-07-10
sleep for longer before starting test

charlesfrye:
The sticky routing test in `server_sticky` is less flaky than it was in `http_server` -- not sure exactly why! but good news.

However, it was still flaking a bit, so I am increasing the sleep time to allow both target containers to scale up.

We are still short of a really excellent test here -- kind of annoying that this feature's behavior is so hard to crisply define. To be rock-solid, the test would need to be statistical and/or based on container counts.

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1612" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


charlesfrye:
@prbot approve

molocule (commented):
lgtm

PR #1621, opened 2026-07-28
Add DPO alignment example (dpo_trl.py)

arjxnt:
## Summary

Adds `06_gpu_and_ml/reinforcement-learning/dpo_trl.py`, a new example that fills a gap in the existing reinforcement-learning coverage: the folder currently has `grpo_trl.py` (code, TRL), `grpo_verl.py` (math, verl), and `learn_math.py` (math, `verifiers` + sandboxed execution) — all GRPO-based rollout/reward-function training. There's no example for **Direct Preference Optimization** (DPO), which trains directly on static `(chosen, rejected)` preference pairs instead of sampling rollouts through an RL loop — simpler, cheaper, and a good fit when you already have preference data rather than a programmatic reward function.

- Trains a LoRA adapter on `Qwen/Qwen2.5-0.5B-Instruct` against [`trl-lib/ultrafeedback_binarized`](https://huggingface.co/datasets/trl-lib/ultrafeedback_binarized) using TRL's `DPOTrainer`.
- Follows the existing `grpo_trl.py` conventions in this folder: pinned container deps, an HF cache Volume + checkpoint Volume, `max_steps=5` default to keep CI fast (`args`/`cmd` frontmatter unnecessary since the function default is already CI-sized), literate-programming prose.
- The model revision hash was pulled from the live HF API rather than guessed. The dataset's actual schema (`chosen`/`rejected` as full conversations sharing a prefix — the "implicit prompt" format, no separate `prompt` column) was verified against `datasets-server` rather than assumed from memory, and the prose was written to describe that accurately.
- `DPOConfig`/`DPOTrainer` keyword arguments (`beta`, `processing_class`, `peft_config`, etc.) were verified against a real `trl==0.28.0` install rather than assumed.

## Monitoring Checklist

- [x] Example is tested by executing with `modal run` (`cmd` in frontmatter targets `::train`)
- [x] Example runs with no arguments (defaults to `max_steps=5`, `max_examples=256` — small enough for CI)
- [x] No third-party local dependencies beyond `modal` (all deps installed inside the container image only)

## Build Stability Checklist

- [x] All container dependencies pinned to at least SemVer minor (`torch==2.7.0`, `trl==0.28.0`, etc.)
- [x] No `latest` tags used
- [x] `python_version="3.11"` specified on the base image
- [x] `ruff check` and `ruff format` pass

## Test plan

- [x] `ruff check` / `ruff format --check` — pass
- [x] `python -m py_compile` — passes
[...]

PR #1623, opened 2026-07-28
Add reward model training example (reward_model_trl.py)

arjxnt:
## Summary

Adds `06_gpu_and_ml/reinforcement-learning/reward_model_trl.py`, filling the last remaining gap in this folder's coverage of preference-based training approaches: **reward model training**, the piece a PPO-style RLHF pipeline needs to actually produce a reward signal (as opposed to `grpo_trl.py`'s programmatic reward function, or `dpo_trl.py`'s [#1621] direct optimization on preference pairs with no separate reward model at all).

- Trains a sequence-classification head (`num_labels=1`) with a LoRA adapter (`task_type="SEQ_CLS"`) on `Qwen/Qwen2.5-0.5B-Instruct`, using TRL's `RewardTrainer`.
- Reuses the same `trl-lib/ultrafeedback_binarized` dataset as `dpo_trl.py` for consistency across the three examples — I confirmed by reading `reward_trainer.py`'s dataset-processing path that `RewardTrainer` accepts the same raw `chosen`/`rejected` conversation columns directly (it applies the tokenizer's chat template and tokenizes internally), so no extra preprocessing was needed.
- Cross-links to and from `grpo_trl.py` and `dpo_trl.py` in the prose, explaining when each of the three approaches (programmatic reward / learned reward model / no reward model at all) is the right fit.
- Every `LoraConfig`/`RewardConfig`/`RewardTrainer` keyword argument used here was verified against a real `trl==0.28.0` install rather than assumed — including confirming `RewardConfig` defaults `bf16=True` (which is why a local no-GPU smoke test needs `use_cpu=True` to construct the config at all; the actual L40S container this runs on doesn't have that problem).

## Monitoring Checklist

- [x] Example is tested by executing with `modal run` (`cmd` in frontmatter targets `::train`)
- [x] Example runs with no arguments (defaults to `max_steps=5`, `max_examples=256` — small enough for CI)
- [x] No third-party local dependencies beyond `modal` (all deps installed inside the container image only)

## Build Stability Checklist

- [x] All container dependencies pinned to at least SemVer minor (`torch==2.7.0`, `trl==0.28.0`, `transformers==4.57.1` — not the yanked `4.57.0`, etc.)
- [x] No `latest` tags used
- [x] `python_version="3.11"` specified on the base image
- [x] `ruff check` and `ruff format` pass

## Test plan

- [x] `ruff check` / `ruff format --check` — pass
- [x] `python -m py_compile` — passes
- [x] `LoraConfig(task_type="SEQ_CLS")` and every `RewardConfig` kwarg verified to construct against a real `trl==0.28.0` install
- [x] Confirmed via source (`reward_trainer.py`) that the dataset's raw `chosen`/`rejected` format is accepted directly by `RewardTrainer`
- [ ] Full `modal run` execution (requires Modal auth I don't have configured locally — this repo's CI runs changed examples automatically on PRs, so happy to have that be the live verification, or for a maintainer to run it)

---

<a href="https://modal.devinenterprise.com/review/modal-labs/modal-examples/pull/1623" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/06_gpu_and_ml/reinforcement-learning/reward_model_trl.py
+++ b/06_gpu_and_ml/reinforcement-learning/reward_model_trl.py
@@ -1,0 +1,152 @@
+# ---
+# cmd: ["modal", "run", "06_gpu_and_ml/reinforcement-learning/reward_model_trl.py::train"]
+# ---
+
+# # Train a reward model for RLHF using TRL
+
+# This example demonstrates how to train a reward model on Modal using TRL's
+# [`RewardTrainer`](https://huggingface.co/docs/trl/main/en/reward_trainer).
+
+# A reward model scores a `(prompt, response)` pair with a single scalar,
+# learned from pairwise human (or AI) preference data. It's the missing piece
+# for classic RLHF pipelines built on PPO: PPO needs a reward *signal* to
+# optimize against, which this example produces, whereas the
+# [GRPO example](https://modal.com/docs/examples/grpo_trl) sidesteps a learned
+# reward model entirely by scoring rollouts with a programmatic reward function,
+# and the [DPO example](https://modal.com/docs/examples/dpo_trl) skips the RL
+# loop altogether and optimizes directly on the preference pairs. All three are
+# valid ways to turn preference data into a better model — which one fits
+# depends on whether you have a programmatic reward, a static preference
+# dataset, or need a reusable scorer for downstream RL.
+
+# ## Setup
+
+# Import the necessary modules for Modal deployment.
+from pathlib import Path
+
+import modal
+
+MINUTES = 60  # seconds
+
+# ## Defining the image and app
+
+app = modal.App("example-reward-model-trl")
+
+# We define an image with TRL, a LoRA-compatible `peft`, and their dependencies.
+# Following the [pinning conventions](https://modal.com/docs/guide/images) for this
+# repo, every container dependency is pinned to at least a SemVer minor version.
+
+image = modal.Image.debian_slim(python_version="3.11").uv_pip_install(
+    "torch==2.7.0",
+    "transformers==4.57.1",  # 4.57.0 is yanked from PyPI
+    "trl==0.28.0",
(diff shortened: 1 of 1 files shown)


arjxnt:
Thanks — went through all three. Fixed one, and verified the other two against the installed `trl==0.28.0` source rather than assuming:

**1. `HF_XET_HIGH_PERFORMANCE` missing** — real miss, fixed in bf02637. Same fix applied to #1621.

**2. `pad_token_id` reliance** — verified safe. `RewardTrainer.__init__` explicitly populates it, in `trl/trainer/reward_trainer.py`:

```python
pad_token = args.pad_token or processing_class.pad_token or processing_class.eos_token   # L407
pad_token_id = processing_class.convert_tokens_to_ids(pad_token)                         # L408
if pad_token_id is None:
    raise ValueError(...)                                                                # L409-413
model.config.pad_token_id = pad_token_id                                                 # L415
processing_class.pad_token_id = pad_token_id                                             # L416
```

So `model.config.pad_token_id` is set before any forward pass, with a three-level fallback ending at `eos_token` (which Qwen2.5's tokenizer does define), and it raises a clear error rather than failing silently if the token isn't in the vocab. The batch-size>1 path is fine.

**3. `center_rewards_coefficient` possibly ignored** — verified it *is* consumed, same file:

```python
if self.args.center_rewards_coefficient is not None:                                          # L586
    loss += self.args.center_rewards_coefficient * torch.mean((rewards_chosen + rewards_rejected) ** 2)  # L587
```

Not silently dropped in the rewrite — the centering penalty is applied to the loss as intended.

Worth noting the review was right that config-construction verification alone wouldn't have caught #3; that's why I went to the loss implementation for both. Full `modal run` still hasn't been executed on my end (no Modal auth configured locally), so CI or a maintainer run remains the real end-to-end check.

charlesfrye:
Hello!

